PR #30, opened 2018-07-09
closes #29

### LINKED ISSUE #29: Calling the size() of a broken CyphertextFile throws IllegalArgumentException
The caused highlevel-error is described here: https://github.com/cryptomator/cryptomator/issues/673

Getting the size of an invalid cyphertext file throws an (undocumented) IllegalArgumentException.
 
We should either document this behavoiur und handle it in the corresponding nio-adapter libraries or should change it e.g. such that a negative size is returned (or something else).

The problem causing code is in CrypoBasicFileAttribute:
https://github.com/cryptomator/cryptofs/blob/04cb671690ae87c5baf03a57a53d1d224a1d404f/src/main/java/org/cryptomator/cryptofs/CryptoBasicFileAttributes.java#L58

infeo:
closes #29

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/CryptoBasicFileAttributes.java
+++ b/src/main/java/org/cryptomator/cryptofs/CryptoBasicFileAttributes.java
@@ -44,6 +44,12 @@ class CryptoBasicFileAttributes implements DelegatingBasicFileAttributes {
 		return false;
 	}
 
+	/**
+	 * Gets the size of the decrypted file.
+	 *
+	 * @return the size of the encrypted file
+	 * @throws IllegalArgumentException if the computed fileSize of a normal file is negative (e.g. it is altered manually)
+	 */
 	@Override
 	public long size() {
 		if (isDirectory()) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/cryptomator/cryptofs/CryptoBasicFileAttributes.java
```diff
@@ -44,6 +44,12 @@ public boolean isSymbolicLink() {
 		return false;
 	}
 
+	/**
+	 * Gets the size of the decrypted file.
+	 *
+	 * @return the size of the encrypted file
+	 * @throws IllegalArgumentException if the computed fileSize of a normal file is negative (e.g. it is altered manually)
```
overheadhunter: to be more specific, I'd suggest:

```
@throws IllegalArgumentException if the delegate reports a size that is considered invalid for a well-formed ciphertext file (e.g. it has been altered manually)
```

src/main/java/org/cryptomator/cryptofs/CryptoBasicFileAttributes.java
```diff
@@ -44,6 +44,12 @@ public boolean isSymbolicLink() {
 		return false;
 	}
 
+	/**
+	 * Gets the size of the decrypted file.
+	 *
+	 * @return the size of the encrypted file
```
overheadhunter: decrypted!!

PR #62, opened 2019-07-11
closes #61

### LINKED ISSUE #61: Backup: Make the key file more robust to modifications
Version: 1.8.6

## Short Description
If the masterkey file is replaced, the backup file should not be overwritten at the next unlock and rather a new one should be created. It is suggested to add a truncated SHA256 value of the corresponding masterkey file to the backup filename.

## Description
cryptofs is creating a backup of the masterkey file in the case the original gets corrupted. This backup is written every time the vault is successfully unlocked. If now the masterkey file gets accidently overwritten (creating a new vault in the existing one), the content of the vault cannot be read anymore and due to the successful unlock also the backup will get overwritten.

The idea is now to link the backup fail directly to a specific masterkey file by adding a truncated SHA 256 value of the masterkey file to the backup name.

The affected code would be :
https://github.com/cryptomator/cryptofs/blob/3ce8537eb84f887d608c3e2d9396308898bccb71/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemModule.java#L28-L34

infeo:
Backup file will now be suffixed with first 7 characters of the SHA256 checksum fo the masterkey file

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemModule.java
+++ b/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemModule.java
@@ -46,7 +46,7 @@ class CryptoFileSystemModule {
 			byte[] digest = md.digest(fileBytes);
 			return BaseEncoding.base16().encode(digest, 0, 4);
 		} catch (NoSuchAlgorithmException e) {
-			throw new IllegalStateException("This Java Platform must support the Message Digest algorithm SHA-256");
+			throw new IllegalStateException("Every Java Platform must support the Message Digest algorithm SHA-256");
 		}
 	}
 
--- a/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemModule.java
+++ b/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemModule.java
@@ -15,6 +15,8 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/cryptomator/cryptofs/CryptoFileSystemModule.java
```diff
@@ -37,6 +39,20 @@ public Cryptor provideCryptor(CryptorProvider cryptorProvider, @PathToVault Path
...
 
+	private String generateFileIdIfPossible(byte[] fileBytes) {
+		try {
+			MessageDigest md = MessageDigest.getInstance("SHA-256");
+			byte[] truncatedDigest = md.digest(fileBytes);
+			StringBuilder sb = new StringBuilder();
+			for (int i = 0; i < 4; i++) { // we only need the first 4 bytes, since each is translated into two characters
+				sb.append(String.format("%02x", truncatedDigest[i]));
+			}
+			return sb.toString().substring(0, 7);
```
overheadhunter: Easier using Guava's [BaseEncoding](https://guava.dev/releases/27.1-jre/api/docs/com/google/common/io/BaseEncoding.html#encode-byte:A-int-int-):

```suggestion
			return BaseEncoding.base16().lowerCase().encode(digest, 0, 4);
```

overheadhunter: And let's just use 8 chars (i.e. 4 bytes) instead of 7. :wink:

src/main/java/org/cryptomator/cryptofs/CryptoFileSystemModule.java
```diff
@@ -37,6 +39,20 @@ public Cryptor provideCryptor(CryptorProvider cryptorProvider, @PathToVault Path
...
+		} catch (NoSuchAlgorithmException e) {
+			return "";
```
overheadhunter:
```suggestion
			throw new IllegalStateException(e);
```

Quoting [MessageDigest JavaDoc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/security/MessageDigest.html):

> Every implementation of the Java platform is required to support the following standard MessageDigest algorithms:
>
> MD5
> SHA-1
> SHA-256

We might even be able to throw an AssertionError, but some static code analysis tools might not like this.

src/main/java/org/cryptomator/cryptofs/CryptoFileSystemModule.java
```diff
@@ -37,6 +39,20 @@ public Cryptor provideCryptor(CryptorProvider cryptorProvider, @PathToVault Path
 		}
 	}
 
+	private String generateFileIdIfPossible(byte[] fileBytes) {
```
overheadhunter: Drop `IfPossible`. Why shouldn't it be possible? See comment below regarding handling of `NoSuchAlgorithmException`.

src/main/java/org/cryptomator/cryptofs/CryptoFileSystemModule.java
```diff
@@ -37,6 +40,16 @@ public Cryptor provideCryptor(CryptorProvider cryptorProvider, @PathToVault Path
 		}
 	}
 
+	private String generateFileId(byte[] fileBytes) {
+		try {
+			MessageDigest md = MessageDigest.getInstance("SHA-256");
+			byte[] digest = md.digest(fileBytes);
+			return BaseEncoding.base16().encode(digest, 0, 4);
+		} catch (NoSuchAlgorithmException e) {
+			throw new IllegalStateException("Every Java Platform must support the Message Digest algorithm SHA-256");
```
overheadhunter:
```suggestion
			throw new IllegalStateException("Every Java Platform must support the Message Digest algorithm SHA-256", e);
```

keep the full strack trace

infeo: put the whole method in an own class and added there the exception in the rethrow

PR #90, opened 2020-09-25
Feature/#88 improve backup strategy

### LINKED ISSUE #88: Masterkey backup strategies: Not all cases covered.
With https://github.com/cryptomator/cryptofs/issues/86, cryptofs can now also unlock vaults located on read-only storage devices.

But the fix is very narrow implemented:
https://github.com/cryptomator/cryptofs/blob/d1be7a7e4cf5c24837556767b4596807c2843947/src/main/java/org/cryptomator/cryptofs/common/MasterkeyBackupHelper.java#L58-L63

Only if the write operation returns  `AccessDeniedException` we check for an alternative. But some file systems are sloppy regarding the specifc errors leading up to just throwing an `IOException` with a message like  "Permission Denied".

Hence we should just catch any IOException and try the workaround. If this one also fails, nothing is lost because the underlying filesystems seems broken and if not, the caller has (maybe only read) access to the encrypted files.

infeo:
This PR closes #88 

The backup strategy is now altered by only throwing an exception, if the masterkey file cannot be read. All other errors (comparsion with backup file, corrupt backup file, etc) are only logged on `WARN` level, but do not throw an exception.

The thought behind it is, that the backup strategy is only a best-effort attempt and for an application it is more important getting access to encrypted data rather than having a backup.

Additionally the method is renamed to better show its functionality.

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/common/MasterkeyBackupHelper.java
+++ b/src/main/java/org/cryptomator/cryptofs/common/MasterkeyBackupHelper.java
@@ -20,7 +20,7 @@ import java.security.NoSuchAlgorithmException;
  * Utility class for generating a suffix for the backup file to make it unique to its original master key file.
  */
 public final class MasterkeyBackupHelper {
-	
+
 	private static final Logger LOG = LoggerFactory.getLogger(MasterkeyBackupHelper.class);
 
 	/**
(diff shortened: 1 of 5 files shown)


src/main/java/org/cryptomator/cryptofs/common/MasterkeyBackupHelper.java
```diff
@@ -47,37 +41,40 @@ public static String generateFileIdSuffix(byte[] fileBytes) {
...
+			LOG.warn("Failed to backup valid masterkey file.");
 		}
 		return backupFilePath;
 	}
-	
-	private static void assertExistingBackupMatchesContent(Path backupFilePath, ByteBuffer expectedContent) throws IOException {
+
+	private static void assertExistingBackupMatchesContent(Path backupFilePath, ByteBuffer expectedContent) {
 		if (Files.exists(backupFilePath)) {
 			// TODO replace by Files.mismatch() when using JDK > 12
```
tobihagemann, line 64: If you're already making changes in that method, maybe this TODO could be resolved? 😄

overheadhunter (approved):

PR #99, opened 2021-03-12
Feature/#98 change vault init

### LINKED ISSUE #98: Exclusively create encrypted file system root at vault initialization
## Description

Currently, everytime a vault is opened/ a cryptofilesystem is created via the CryptoFileSystemProvider, the encrypted root directory is created in the constructor of the CryptoFileSystemImpl:
```Java
	public CryptoFileSystemImpl(CryptoFileSystemProvider provider, CryptoFileSystems cryptoFileSystems /*, ... more arguments */ ) {
		/* a lot assingments */

		rootDirectoryInitializer.initialize(rootPath);
	}
```

The class `RootDirectoryInitializer` does nothing else except this.

This design decison is bad, because the file system root directory is as the main entry point an inherent part of vault and not subject of change. As such, if it is _not_ present, an error should be thrown. Just recreating the root is not a fix, beause the stored data is already lost and additionally it suggests that there is no error at all.

## Suggested Solution
Only create the encrypted vault root at initialization, i.e. i the [`CryptoFileSystemProvider.initiatlize(...)` method](https://github.com/cryptomator/cryptofs/blob/de4d0faee36f42acdfc2b956be34cfbba87f7bb1/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemProvider.java#L127). Remove the `RootDirectoryInitializer` class.

infeo:
Closes #98 

## Implemented Solution
With this PR, the content root of a vault will be exclusivley created during the vault initialization, i.e. in the _static_ `CryptoFileSystemProvider.initialize(Path vaultPath)` method and never implicitily during CryptoFileSystem(Impl) creation.

Additionally, when creating a new CryptoFileSystem in the `CryptoFileSystems` class, the existence of the vault content root is checked before the the filesystem is actually constructed and an IOException raised, if the check fails.


src/main/java/org/cryptomator/cryptofs/CryptoFileSystems.java
```diff
@@ -54,9 +54,11 @@ public CryptoFileSystemImpl create(CryptoFileSystemProvider provider, Path pathT
 			rawKey = key.getEncoded();
 			var config = configLoader.verify(rawKey, Constants.VAULT_VERSION);
 			var adjustedProperties = adjustForCapabilities(pathToVault, properties);
+			Cryptor cryptor = config.getCipherCombo().getCryptorProvider(csprng).withKey(key);
```
overheadhunter, line 57: I wonder, how the masterkey survives this method... It will be destroyed when leavin the try-with-resource block. Thus the cryptor should no longer be useful... 🤔

src/main/java/org/cryptomator/cryptofs/CryptoFileSystems.java
```diff
@@ -66,9 +68,22 @@ public CryptoFileSystemImpl create(CryptoFileSystemProvider provider, Path pathT
...
 
+	/**
+	 * Checks if the vault has a content root folder. If not, an exception is raised.
+	 * @param pathToVault Path to the vault root
+	 * @param cryptor Cryptor object initialized with the correct masterkey
+	 * @throws ContentRootMissingException If the existence of encrypted vault content root cannot be ensured
+	 */
+	private void checkVaultRootExistence(Path pathToVault, Cryptor cryptor) throws ContentRootMissingException {
+		String dirHash = cryptor.fileNameCryptor().hashDirectoryId(Constants.ROOT_DIR_ID);
+		Path vaultCipherRootPath = pathToVault.resolve(Constants.DATA_DIR_NAME).resolve(dirHash.substring(0, 2)).resolve(dirHash.substring(2));
```
overheadhunter, line 79: This `pathToVault.resolve(Constants.DATA_DIR_NAME).resolve(dirHash.substring(0, 2)).resolve(dirHash.substring(2))` should be deduplicated across the project (in a different PR)

src/main/java/org/cryptomator/cryptofs/CryptoFileSystems.java
```diff
@@ -66,9 +68,22 @@ public CryptoFileSystemImpl create(CryptoFileSystemProvider provider, Path pathT
...
+		if(! Files.exists(vaultCipherRootPath)){
```
overheadhunter, line 80: reformat

src/main/java/org/cryptomator/cryptofs/CryptoFileSystems.java
```diff
@@ -66,9 +68,22 @@ public CryptoFileSystemImpl create(CryptoFileSystemProvider provider, Path pathT
...
+			throw new ContentRootMissingException("The encrypted root directory of the vault "+pathToVault+" is missing.");
```
overheadhunter, line 81: reformat

overheadhunter (changes requested):

src/main/java/org/cryptomator/cryptofs/CryptoFileSystems.java
infeo: Resolved with 7b4213d

src/main/java/org/cryptomator/cryptofs/CryptoFileSystems.java
infeo: Resolved with 7b4213d

src/main/java/org/cryptomator/cryptofs/CryptoFileSystems.java
infeo: Will be discussed in https://github.com/cryptomator/cryptolib/issues/19

src/main/java/org/cryptomator/cryptofs/CryptoFileSystems.java
```diff
@@ -54,21 +54,41 @@ public CryptoFileSystemImpl create(CryptoFileSystemProvider provider, Path pathT
...
 			var config = configLoader.verify(rawKey, Constants.VAULT_VERSION);
 			var adjustedProperties = adjustForCapabilities(pathToVault, properties);
-			return fileSystems.compute(normalizedPathToVault, (path, fs) -> {
-				if (fs == null) {
-					return create(provider, normalizedPathToVault, adjustedProperties, key, config);
-				} else {
-					throw new FileSystemAlreadyExistsException();
-				}
-			});
+			var cryptor = config.getCipherCombo().getCryptorProvider(csprng).withKey(key);
```
overheadhunter, line 57: Until there is new API related to cryptomator/cryptolib#19, make a copy, so the destruction of `key` at the end of this method doesn't void the cryptor.

```suggestion
			var keyCopy = Masterkey.createFromRaw(key.getEncoded()); // TODO replace with key.clone() eventually (depends on cryptomator/cryptolib#19
			var cryptor = config.getCipherCombo().getCryptorProvider(csprng).withKey(keyCopy); 

```

overheadhunter (changes requested):

src/main/java/org/cryptomator/cryptofs/CryptoFileSystems.java
overheadhunter: See #100

overheadhunter (approved):

PR #104, opened 2021-05-05
Feature/#103 refactor contains vault

### LINKED ISSUE #103: CryptoFileSystemProvider.containsVault(path) returns true for invalid vault
If a version 8 vault contains the masterkey file and the `d` directory, but the config file is missing, `CryptoFileSystemProvider.containsVault(...)` returns true:

https://github.com/cryptomator/cryptofs/blob/fe410fdef30bf59bc007fb56828501a6439423fc/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemProvider.java#L171-L176

From my point of view, this function should check the given path only with the criteria of the most recent vault version. At the time of opening this ticket, this would be vault format 8 and the actual check would a existence test of the vault config file and the data directory.

A consequence of this would be, that testing a vault of a lower version would return false. But that's ok, for this we have the migration package.

infeo:
Closes #103 

As suggested, the (public) enum `DirStructure` is introduced with a static method to determine how "much" the directory structure resembles a vault.

I think the most important change is that the `containsVault(...)` method only checks for a vault structure of vault format 8. For a more fine grained result, use the method `CryptoFileSystemProvider.checkDirStructureForVault(...)`.

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemProvider.java
+++ b/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemProvider.java
@@ -172,7 +172,7 @@ public class CryptoFileSystemProvider extends FileSystemProvider {
 	 */
 	public static boolean containsVault(Path pathToAssumedVault, String vaultConfigFilename, String masterkeyFilename) {
 		try {
-			return checkDirStructure(pathToAssumedVault, vaultConfigFilename, masterkeyFilename) == DirStructure.VAULT;
+			return checkDirStructureForVault(pathToAssumedVault, vaultConfigFilename, masterkeyFilename) == DirStructure.VAULT;
 		} catch (IOException e) {
 			return false;
 		}
@@ -188,7 +188,7 @@ public class CryptoFileSystemProvider extends FileSystemProvider {
 	 * @throws IOException
 	 * @since 2.0.0
 	 */
-	public static DirStructure checkDirStructure(Path pathToAssumedVault, String vaultConfigFilename, String masterkeyFilename) throws IOException {
+	public static DirStructure checkDirStructureForVault(Path pathToAssumedVault, String vaultConfigFilename, String masterkeyFilename) throws IOException {
 		return DirStructure.checkDirStructure(pathToAssumedVault, vaultConfigFilename, masterkeyFilename);
 	}
 
--- a/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemProvider.java
+++ b/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemProvider.java
@@ -161,18 +161,35 @@ public class CryptoFileSystemProvider extends FileSystemProvider {
 
 	/**
 	 * Checks if the folder represented by the given path exists and contains a valid vault structure.
+	 * <p>
+	 * See {@link DirStructure#VAULT} for the criterias of being a valid vault.
 	 *
-	 * @param pathToVault         A directory path
+	 * @param pathToAssumedVault A directory path
 	 * @param vaultConfigFilename Name of the vault config file
-	 * @param masterkeyFilename   Name of the masterkey file
+	 * @param masterkeyFilename Name of the masterkey file
 	 * @return <code>true</code> if the directory seems to contain a vault.
 	 * @since 2.0.0
 	 */
-	public static boolean containsVault(Path pathToVault, String vaultConfigFilename, String masterkeyFilename) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/cryptomator/cryptofs/DirStructure.java
```diff
@@ -0,0 +1,69 @@
...
+import java.nio.file.NotDirectoryException;
+import java.nio.file.Path;
+import java.nio.file.attribute.BasicFileAttributes;
+
+/**
+ * Enumeration of the vault directory structure resemblances.
+ * <p>
+ * A valid vault must contain a `d` directory.
+ * If the vault version is 8, it must also contains a vault config file.
+ * If the vault version is smaller than 8, it must also contain a masterkey file.
```
overheadhunter, line 16:
```suggestion
 * Beginning with vault format 8, it must also contain a vault config file.
 * If the vault format is lower than 8, it must instead contain a masterkey file.
```

src/main/java/org/cryptomator/cryptofs/DirStructure.java
```diff
@@ -0,0 +1,69 @@
...
+ * <p>
+ * In the latter case, to distinct between a damaged vault 8 directory and a legacy vault the masterkey file must be read.
```
overheadhunter, line 18:
```suggestion
 * In the latter case, to distinguish between a damaged vault 8 directory and a legacy vault the masterkey file must be read.
```

src/main/java/org/cryptomator/cryptofs/CryptoFileSystemProvider.java
```diff
@@ -161,18 +161,35 @@ public static void initialize(Path pathToVault, CryptoFileSystemProperties prope
...
+	 * @param masterkeyFilename Name of the masterkey file
 	 * @return <code>true</code> if the directory seems to contain a vault.
 	 * @since 2.0.0
 	 */
-	public static boolean containsVault(Path pathToVault, String vaultConfigFilename, String masterkeyFilename) {
-		Path vaultConfigPath = pathToVault.resolve(vaultConfigFilename);
-		Path masterkeyPath = pathToVault.resolve(masterkeyFilename);
-		Path dataDirPath = pathToVault.resolve(Constants.DATA_DIR_NAME);
-		return (Files.isReadable(vaultConfigPath) || Files.isReadable(masterkeyPath)) && Files.isDirectory(dataDirPath);
+	public static boolean containsVault(Path pathToAssumedVault, String vaultConfigFilename, String masterkeyFilename) {
```
overheadhunter, line 173: Let's use the opportunity to ditch this method. No need to retain it, we have already removed a lot of other deprecated methods in 2.0.0

src/main/java/org/cryptomator/cryptofs/DirStructure.java
```diff
@@ -0,0 +1,69 @@
...
+	 * Analyzes the structure of the given directory under certain vault existence criteria.
+	 *
+	 * @param pathToVault A directory path
+	 * @param vaultConfigFilename Name of the vault config file
+	 * @param masterkeyFilename Name of the masterkey file
+	 * @return enum indicating what this directory might be
+	 * @throws IOException if the provided path is not a directory, does not exist or cannot be read
+	 */
+	public static DirStructure checkDirStructure(Path pathToVault, String vaultConfigFilename, String masterkeyFilename) throws IOException {
+		if(! Files.readAttributes(pathToVault, BasicFileAttributes.class).isDirectory()) {
```
overheadhunter, line 54:
```suggestion
		if(!Files.isDirectory(pathToVault)) {
```

overheadhunter (changes requested):
I am even considering removing _both_ methods from CryptoFileSystemProvider. After all, the new enum is public.

src/main/java/org/cryptomator/cryptofs/DirStructure.java
infeo: reading attributes is by intent such that a `NoSuchFileException` can be thrown.Otherwise this this excpetion is swallowed by `Files.isDirectory()`.

src/main/java/org/cryptomator/cryptofs/CryptoFileSystemProvider.java
infeo: Done in c92a45c

infeo:
For useability I would keep the `CryptoFileSystemProvider.checkDirStructureForVault(...)` method. It just a delegate, but we gather vault structure related functions (initialize, check) at one place.

overheadhunter (approved):

PR #105, opened 2021-06-09
Extend HealthAPI to get more specific information about results

infeo:
The `DiagnosticResult` interface has a new method:
```Java
/**
 * Get more specific info about the result like names of affected resources.
 *
 * @return A map of strings containing result specific information
 */
default Map<String, String> details() {
	return Map.of();
}
```

For all already existing ResultImpl the method is overridden providing specific info.


overheadhunter (approved):
👍

PR #106, opened 2021-06-14
Health-API: Implement OrphanDir.fix()

infeo:
This PR implements the `OrphanDir::fix` method.

The fix is implemented in a way that it restricts itself only to fixing the orphan dir problem. Other problems like a missing or too big `dir.c9r` are ignored. Hence, the fix must preserve the shortend property of a ciphertext file.

To get an understanding, consider an orphaned ciphertext dir with a shortened file, where `content.c9r` is missing:
```
-d/NC/RWII6WWFYQTWWEMBXK4OOLUMUHT6PX
    |- 7HkyoIUdCinG8NxXbwD5hD6nNsE=.c9s
        |- name.c9s
        // missing content.c9s
```

Since the content.c9s is missing, the file cannot be simply unshortend. By keeping the file shortened, it is ensured that the orphaned directory can be integrated into the vault structure without stopping at these errors.

All files inside the orphaned directory will be moved to a directory with the cleartext path `/CRYPTOMATOR_RECOVERY/[DirId-Hash of Orphan]`. The name of the "deorphaned"  dir is deterministic, such that after an interrupted fix, a new result will continue placing all orphan files into the same deorphaned directory.
To prevent `FileAlreadyExistsExceptions`, each moved resource in the same run is suffixed with a runId.

`/CRYPTOMATOR_RECOVERY` is a new directory with a unique dir id. All "deorphaned" directories will be located their. It is only created on demand by the fix() method. From the user perspective it will be a normal directory. If a user creates in the vault content root a directory with this name, it won't have the unique id. Every orphanDir fix will fail and the solution is to rename the user generated dir. This can be implemetned as an additonal check or result. If the recovery dir is orphaned itself, just the missing c9r-directory and dirId-file are created and not files moved. An orphaned, user created directory with the same name will stay orphaned and is handled via standard deorphan procedure.

In this context, the Health API is adjusted, that  `fix(...)` throws an IOException-

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/common/Constants.java
+++ b/src/main/java/org/cryptomator/cryptofs/common/Constants.java
@@ -17,7 +17,8 @@ public final class Constants {
 	public static final String MASTERKEY_BACKUP_SUFFIX = ".bkup";
 	public static final String DATA_DIR_NAME = "d";
 	public static final String ROOT_DIR_ID = "";
-	public static final String RECOVERY_DIR_ID = "l+f";
+	public static final String RECOVERY_DIR_ID = "recovery";
+
 	public static final String CRYPTOMATOR_FILE_SUFFIX = ".c9r";
 	public static final String DEFLATED_FILE_SUFFIX = ".c9s";
 	public static final String DIR_FILE_NAME = "dir.c9r";
--- a/src/main/java/org/cryptomator/cryptofs/common/Constants.java
+++ b/src/main/java/org/cryptomator/cryptofs/common/Constants.java
@@ -17,6 +17,7 @@ public final class Constants {
 	public static final String MASTERKEY_BACKUP_SUFFIX = ".bkup";
 	public static final String DATA_DIR_NAME = "d";
 	public static final String ROOT_DIR_ID = "";
+	public static final String RECOVERY_DIR_ID = "l+f";
 	public static final String CRYPTOMATOR_FILE_SUFFIX = ".c9r";
 	public static final String DEFLATED_FILE_SUFFIX = ".c9s";
 	public static final String DIR_FILE_NAME = "dir.c9r";
@@ -29,4 +30,5 @@ public final class Constants {
 	public static final int MAX_DIR_FILE_LENGTH = 36; // UUIDv4: hex-encoded 16 byte int + 4 hyphens = 36 ASCII chars
 
 	public static final String SEPARATOR = "/";
+	public static final String RECOVERY_DIR_NAME = "CRYPTOMATOR_RECOVERY";
 }
--- a/src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
@@ -119,9 +119,14 @@ public class OrphanDir implements DiagnosticResult {
 		//create "step-parent" directory to move orphaned files to
 		String cipherStepParentDirName = convertClearToCiphertext(cryptor, clearStepParentDirName, Constants.RECOVERY_DIR_ID);
 		Path cipherStepParentDirFile = cipherRecoveryDir.resolve(cipherStepParentDirName + "/" + Constants.DIR_FILE_NAME);
-		Files.createDirectory(cipherStepParentDirFile.getParent());
-		var stepParentUUID = UUID.randomUUID().toString();
-		Files.writeString(cipherStepParentDirFile, stepParentUUID, StandardCharsets.UTF_8, StandardOpenOption.CREATE_NEW, StandardOpenOption.WRITE);
+		final String stepParentUUID;
+		if (Files.exists(cipherStepParentDirFile, LinkOption.NOFOLLOW_LINKS)) {
+			stepParentUUID = Files.readString(cipherStepParentDirFile, StandardCharsets.UTF_8);
+		} else {
+			Files.createDirectory(cipherStepParentDirFile.getParent());
+			stepParentUUID = UUID.randomUUID().toString();
+			Files.writeString(cipherStepParentDirFile, stepParentUUID, StandardCharsets.UTF_8, StandardOpenOption.CREATE_NEW, StandardOpenOption.WRITE);
+		}
 		String stepParentDirHash = cryptor.hashDirectoryId(stepParentUUID);
 		Path stepParentDir = dataDir.resolve(stepParentDirHash.substring(0, 2)).resolve(stepParentDirHash.substring(2)).toAbsolutePath();
 		Files.createDirectories(stepParentDir);
--- a/src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
@@ -65,7 +65,7 @@ public class OrphanDir implements DiagnosticResult {
 		Path orphanedDir = pathToVault.resolve(Constants.DATA_DIR_NAME).resolve(this.dir);
 		String orphanHash = dir.getParent().getFileName().toString() + dir.getFileName().toString();
 
-		var stepParentDir = prepareCryptoFilesystem(pathToVault, cryptor.fileNameCryptor(), orphanHash);
+		var stepParentDir = prepareStepParent(pathToVault, cryptor.fileNameCryptor(), orphanHash);
 
 		try (var orphanedContentStream = Files.newDirectoryStream(orphanedDir)) {
 			AtomicInteger fileCounter = new AtomicInteger(1);
@@ -86,7 +86,7 @@ public class OrphanDir implements DiagnosticResult {
 	}
 
 	// visible for testing
-	CryptoPathMapper.CiphertextDirectory prepareCryptoFilesystem(Path pathToVault, FileNameCryptor cryptor, String clearStepParentDirName) throws IOException {
+	CryptoPathMapper.CiphertextDirectory prepareStepParent(Path pathToVault, FileNameCryptor cryptor, String clearStepParentDirName) throws IOException {
 		//determine path for cipher root
 		Path dataDir = pathToVault.resolve(Constants.DATA_DIR_NAME);
 		String rootDirHash = cryptor.hashDirectoryId(Constants.ROOT_DIR_ID);
--- a/src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
@@ -80,7 +80,7 @@ public class OrphanDir implements DiagnosticResult {
 			AtomicInteger fileCounter = new AtomicInteger(1);
 			AtomicInteger dirCounter = new AtomicInteger(1);
 			AtomicInteger symlinkCounter = new AtomicInteger(1);
-			String veryLongSuffix = createClearnameToBeShortened(config.getShorteningThreshold());
+			String longNameSuffix = createClearnameToBeShortened(config.getShorteningThreshold());
 
 			for (Path orphanedResource : orphanedContentStream) {
 				var newClearName = switch (determineCiphertextFileType(orphanedResource)) {
@@ -88,22 +88,7 @@ public class OrphanDir implements DiagnosticResult {
 					case DIRECTORY -> DIR_PREFIX + dirCounter.getAndIncrement();
 					case SYMLINK -> SYMLINK_PREFIX + symlinkCounter.getAndIncrement();
 				};
-				if (orphanedResource.toString().endsWith(Constants.DEFLATED_FILE_SUFFIX)) {
-					var newCipherName = convertClearToCiphertext(cryptor.fileNameCryptor(), newClearName + veryLongSuffix, stepParentDir.dirId);
-					var deflatedName = BaseEncoding.base64Url().encode(SHA1_HASHER.digest(newCipherName.getBytes(StandardCharsets.UTF_8))) + Constants.DEFLATED_FILE_SUFFIX;
-					Path targetPath = stepParentDir.path.resolve(deflatedName);
-					Files.move(orphanedResource, targetPath, StandardCopyOption.ATOMIC_MOVE);
-
-					//adjust name.c9s
-					try (var fc = Files.newByteChannel(targetPath.resolve(Constants.INFLATED_FILE_NAME), StandardOpenOption.WRITE, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)) {
-						fc.write(ByteBuffer.wrap(newCipherName.getBytes(StandardCharsets.UTF_8)));
-					}
-				} else {
-					var newCipherName = convertClearToCiphertext(cryptor.fileNameCryptor(), newClearName, stepParentDir.dirId);
-					Path targetPath = stepParentDir.path.resolve(newCipherName);
-					Files.move(orphanedResource, targetPath, StandardCopyOption.ATOMIC_MOVE);
-				}
-
+				adoptOrphanedResource(new Adoption(newClearName, orphanedResource), stepParentDir, cryptor.fileNameCryptor(), longNameSuffix);
 			}
 		}
 		Files.delete(orphanedDir);
@@ -143,6 +128,27 @@ public class OrphanDir implements DiagnosticResult {
 		return new CryptoPathMapper.CiphertextDirectory(stepParentUUID, stepParentDir);
 	}
 
+	// visible for testing
+	void adoptOrphanedResource(Adoption adoption, CryptoPathMapper.CiphertextDirectory stepParentDir, FileNameCryptor cryptor, String longNameSuffix) throws IOException {
+		if (adoption.oldCipherPath.toString().endsWith(Constants.DEFLATED_FILE_SUFFIX)) {
+			var newCipherName = convertClearToCiphertext(cryptor, adoption.newClearname + longNameSuffix, stepParentDir.dirId);
+			var deflatedName = BaseEncoding.base64Url().encode(SHA1_HASHER.digest(newCipherName.getBytes(StandardCharsets.UTF_8))) + Constants.DEFLATED_FILE_SUFFIX;
+			Path targetPath = stepParentDir.path.resolve(deflatedName);
+			Files.move(adoption.oldCipherPath, targetPath, StandardCopyOption.ATOMIC_MOVE);
+
+			//adjust name.c9s
+			try (var fc = Files.newByteChannel(targetPath.resolve(Constants.INFLATED_FILE_NAME), StandardOpenOption.WRITE, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)) {
+				fc.write(ByteBuffer.wrap(newCipherName.getBytes(StandardCharsets.UTF_8)));
+			}
+		} else {
+			var newCipherName = convertClearToCiphertext(cryptor, adoption.newClearname, stepParentDir.dirId);
+			Path targetPath = stepParentDir.path.resolve(newCipherName);
+			Files.move(adoption.oldCipherPath, targetPath, StandardCopyOption.ATOMIC_MOVE);
+		}
+	}
+
+	// visible for testing
+	record Adoption(String newClearname, Path oldCipherPath) {}
 
 	private String createClearnameToBeShortened(int threshold) {
 		int neededLength = threshold / 4 * 3 - 16;
--- a/src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
@@ -81,18 +81,17 @@ public class OrphanDir implements DiagnosticResult {
 			String filePrefix = "file";
 			String dirPrefix = "directory";
 			String symlinkPrefix = "symlink";
-			String veryLongSuffix = clearnameToBeDefinitelyShortend(cryptor, config.getShorteningThreshold());
+			String veryLongSuffix = clearnameToBeDefinitelyShortend(config.getShorteningThreshold());
 			MessageDigest sha1Hasher = MessageDigest.getInstance("SHA-1");
 
 			for (Path orphanedResource : orphanedContentStream) {
+				var newClearName = switch (determineCiphertextFileType(orphanedResource)) {
+					case FILE -> filePrefix + fileCounter.getAndIncrement();
+					case DIRECTORY -> dirPrefix + dirCounter.getAndIncrement();
+					case SYMLINK -> symlinkPrefix + symlinkCounter.getAndIncrement();
+				};
 				if (orphanedResource.toString().endsWith(Constants.DEFLATED_FILE_SUFFIX)) {
-					var newClearName = switch (determineCiphertextFileType(orphanedResource)) {
-						case FILE -> filePrefix + fileCounter.getAndIncrement();
-						case DIRECTORY -> dirPrefix + dirCounter.getAndIncrement();
-						case SYMLINK -> symlinkPrefix + symlinkCounter.getAndIncrement();
-					} + veryLongSuffix;
-					var newCipherName = convertClearToCiphertext(cryptor, newClearName, cipherTargetDir.dirId) + Constants.CRYPTOMATOR_FILE_SUFFIX;
-					//deflate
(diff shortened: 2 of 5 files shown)


src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -114,32 +102,40 @@ public void fix(Path pathToVault, VaultConfig config, Masterkey masterkey, Crypt
...
 
+	CryptoPathMapper.CiphertextDirectory prepareCryptoFilesystem(Path pathToVault, FileNameCryptor cryptor, String clearStepParentDirName) throws IOException {
+		Path dataDir = pathToVault.resolve(Constants.DATA_DIR_NAME);
+		String rootDirHash = cryptor.hashDirectoryId(Constants.ROOT_DIR_ID);
+		Path vaultCipherRootPath = dataDir.resolve(rootDirHash.substring(0, 2)).resolve(rootDirHash.substring(2)).toAbsolutePath();
+
+		//create  if not existent with constant id
+		String cipherRecoveryDirName = convertClearToCiphertext(cryptor, Constants.RECOVERY_DIR_NAME, Constants.ROOT_DIR_ID);
+		Path cipherRecoveryDirFile = vaultCipherRootPath.resolve(cipherRecoveryDirName + "/" + Constants.DIR_FILE_NAME);
+		Files.createDirectory(cipherRecoveryDirFile.getParent());
```
overheadhunter, line 113: This will probably fail if it already exists.

src/main/java/org/cryptomator/cryptofs/common/Constants.java
```diff
@@ -17,6 +17,7 @@ private Constants() {
 	public static final String MASTERKEY_BACKUP_SUFFIX = ".bkup";
 	public static final String DATA_DIR_NAME = "d";
 	public static final String ROOT_DIR_ID = "";
+	public static final String RECOVERY_DIR_ID = "l+f";
```
overheadhunter, line 20:
```suggestion
	public static final String RECOVERY_DIR_ID = "recovery";

```

overheadhunter (changes requested):

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
overheadhunter: This case should also be covered by a unit test

infeo: Fixed in 1d0be7d

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -34,4 +68,104 @@ public String toString() {
...
+			Path targetPath = stepParentDir.path.resolve(newCipherName);
+			Files.move(adoption.oldCipherPath, targetPath, StandardCopyOption.ATOMIC_MOVE);
+		}
+	}
+
+	// visible for testing
+	record Adoption(String newClearname, Path oldCipherPath) {}
+
+	private String createClearnameToBeShortened(int threshold) {
+		int neededLength = threshold / 4 * 3 - 16;
```
overheadhunter, line 154: I forgot we need to subtract `.c9r` → 4 chars from the ciphertext name as well. See also:

https://github.com/cryptomator/cryptofs/blob/6445f1ddb06c201dfa6e4e9c81459201c17dec70/src/main/java/org/cryptomator/cryptofs/common/FileSystemCapabilityChecker.java#L99-L101

infeo: Fixed in 6af26a6

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -12,6 +32,20 @@
  */
 public class OrphanDir implements DiagnosticResult {
 
+	private static final String FILE_PREFIX = "file";
+	private static final String DIR_PREFIX = "directory";
+	private static final String SYMLINK_PREFIX = "symlink";
+	private static final String LONG_NAME_SUFFIX_BASE = "_withVeryLongName";
+	private static final MessageDigest SHA1_HASHER;
```
overheadhunter, line 39: This class is not threadsafe. Create a new one within `fix(...)`.

infeo: Fixed in 833d613

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -173,4 +165,12 @@ private CiphertextFileType determineCiphertextFileType(Path ciphertextPath) {
 			return CiphertextFileType.FILE;
 		}
 	}
+
+	private static MessageDigest getSha1MessageDigest() {
+		try {
+			return MessageDigest.getInstance("SHA1");
+		} catch (NoSuchAlgorithmException e) {
+			throw new InstantiationError("Every JVM needs to provide a SHA1 implementation.");
```
overheadhunter, line 173: `InstantiationError` is misleading. While I think that throwing an Error is legitimate in a broken JVM, static code analysis still prefers it if we don't touch them. Throw an `IllegalStateException` instead.

infeo: Fixed in 06d7933

src/main/java/org/cryptomator/cryptofs/common/Constants.java
infeo: resolved in fd5743f

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -33,5 +58,119 @@ public String toString() {
...
+					case SYMLINK -> SYMLINK_PREFIX + symlinkCounter.getAndIncrement();
+				} + "_" + runId;
+				adoptOrphanedResource(new Adoption(newClearName, orphanedResource), stepParentDir, cryptor.fileNameCryptor(), longNameSuffix, sha1);
+			}
+		}
+		Files.delete(orphanedDir);
+	}
+
+	// visible for testing
+	CryptoPathMapper.CiphertextDirectory prepareCryptoFilesystem(Path pathToVault, FileNameCryptor cryptor, String clearStepParentDirName) throws IOException {
```
overheadhunter, line 89:
```suggestion
	CryptoPathMapper.CiphertextDirectory prepareStepParent(Path pathToVault, FileNameCryptor cryptor, String clearStepParentDirName) throws IOException {
```

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -33,5 +58,119 @@ public String toString() {
 		return Map.of(ENCRYPTED_PATH, dir.toString());
 	}
 
-	// fix: create new dirId inside of L+F dir and rename existing dir accordingly.
+	@Override
+	public void fix(Path pathToVault, VaultConfig config, Masterkey masterkey, Cryptor cryptor) throws IOException {
+		var sha1 = getSha1MessageDigest();
+		String runId = BaseEncoding.base64Url().encode(sha1.digest(UUID.randomUUID().toString().getBytes(StandardCharsets.UTF_8))).substring(0, 3);
```
overheadhunter, line 64: no need to sha1 the UUID
```suggestion
		String runId = Integer.toString((short) UUID.randomUUID().getMostSignificantBits(), 32);
```

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -33,5 +58,119 @@ public String toString() {
...
+			}
+		} else {
+			var newCipherName = convertClearToCiphertext(cryptor, adoption.newClearname, stepParentDir.dirId);
+			Path targetPath = stepParentDir.path.resolve(newCipherName);
+			Files.move(adoption.oldCipherPath, targetPath, StandardCopyOption.ATOMIC_MOVE);
+		}
+	}
+
+	// visible for testing
+	record Adoption(String newClearname, Path oldCipherPath) {}
```
overheadhunter, line 148: Why not just have two params in `adoptOrphanedResource`? record isn't used elsewhere, is not added to any collection or stored otherwise. Just having two params is less verbose than creating new tuples all the time.

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -33,5 +58,119 @@ public String toString() {
...
+
+		//check if recovery dir exists and has unique recovery id
+		String cipherRecoveryDirName = convertClearToCiphertext(cryptor, Constants.RECOVERY_DIR_NAME, Constants.ROOT_DIR_ID);
+		Path cipherRecoveryDirFile = vaultCipherRootPath.resolve(cipherRecoveryDirName + "/" + Constants.DIR_FILE_NAME);
+		if (Files.notExists(cipherRecoveryDirFile, LinkOption.NOFOLLOW_LINKS)) {
+			Files.createDirectory(cipherRecoveryDirFile.getParent());
+			Files.writeString(cipherRecoveryDirFile, Constants.RECOVERY_DIR_ID, StandardCharsets.UTF_8, StandardOpenOption.CREATE_NEW, StandardOpenOption.WRITE);
+		} else {
+			String uuid = Files.readString(cipherRecoveryDirFile, StandardCharsets.UTF_8);
+			if (!uuid.equals(Constants.RECOVERY_DIR_ID)) {
```
overheadhunter, line 103: NPE-safe:

```suggestion
			if (!Constants.RECOVERY_DIR_ID.equals(uuid)) {
```

src/test/java/org/cryptomator/cryptofs/health/dirid/OrphanDirTest.java
```diff
@@ -0,0 +1,304 @@
...
+			uuidClass.when(() -> UUID.randomUUID()).thenReturn(uuid);
+			Mockito.doReturn("aaaaaa").when(uuid).toString();
+
+			result.prepareCryptoFilesystem(pathToVault, cryptor, clearStepParentName);
+		}
+
+		Assertions.assertEquals(Constants.RECOVERY_DIR_ID, Files.readString(pathToVault.resolve("d/00/0000/1.c9r/dir.c9r"), StandardCharsets.UTF_8));
+		Assertions.assertEquals("aaaaaa", Files.readString(pathToVault.resolve("d/11/1111/2.c9r/dir.c9r"), StandardCharsets.UTF_8));
+		Assertions.assertTrue(Files.isDirectory(pathToVault.resolve("d/22/2222")));
+	}
```
overheadhunter, line 170: These tests share a lot of test-setup code. Maybe use a `@Nested` test with a common `@BeforeEach`?

overheadhunter (changes requested):
Just a few code improvements

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
overheadhunter: Rename applies to test methods and their display names as well

src/test/java/org/cryptomator/cryptofs/health/dirid/OrphanDirTest.java
infeo: Fixed in 277c51e

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
infeo: Fixed in 2934f3f

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
infeo: superseeded with 420e200

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -108,45 +110,47 @@ public void fix(Path pathToVault, VaultConfig config, Masterkey masterkey, Crypt
 		Path cipherRecoveryDir = dataDir.resolve(recoveryDirHash.substring(0, 2)).resolve(recoveryDirHash.substring(2)).toAbsolutePath();
 		Files.createDirectories(cipherRecoveryDir);
 
+		return cipherRecoveryDir;
+	}
+
+	// visible for testing
+	CryptoPathMapper.CiphertextDirectory prepareStepParent(Path cipherRecoveryDir, FileNameCryptor cryptor, String clearStepParentDirName) throws IOException {
```
overheadhunter, line 117: 👍

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -108,45 +110,47 @@ public void fix(Path pathToVault, VaultConfig config, Masterkey masterkey, Crypt
...
 			stepParentUUID = Files.readString(cipherStepParentDirFile, StandardCharsets.UTF_8);
 		} else {
-			Files.createDirectory(cipherStepParentDirFile.getParent());
+			Files.createDirectories(cipherStepParentDirFile.getParent());
 			stepParentUUID = UUID.randomUUID().toString();
 			Files.writeString(cipherStepParentDirFile, stepParentUUID, StandardCharsets.UTF_8, StandardOpenOption.CREATE_NEW, StandardOpenOption.WRITE);
 		}
 		String stepParentDirHash = cryptor.hashDirectoryId(stepParentUUID);
-		Path stepParentDir = dataDir.resolve(stepParentDirHash.substring(0, 2)).resolve(stepParentDirHash.substring(2)).toAbsolutePath();
+		Path stepParentDir = cipherRecoveryDir.getParent().getParent().resolve(stepParentDirHash.substring(0, 2)).resolve(stepParentDirHash.substring(2)).toAbsolutePath();
```
overheadhunter: revert to datadir, more args are ok

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -63,31 +63,33 @@ public void fix(Path pathToVault, VaultConfig config, Masterkey masterkey, Crypt
...
-				var newClearName = switch (determineCiphertextFileType(orphanedResource)) {
-					case FILE -> FILE_PREFIX + fileCounter.getAndIncrement();
-					case DIRECTORY -> DIR_PREFIX + dirCounter.getAndIncrement();
-					case SYMLINK -> SYMLINK_PREFIX + symlinkCounter.getAndIncrement();
-				} + "_" + runId;
-				adoptOrphanedResource(new Adoption(newClearName, orphanedResource), stepParentDir, cryptor.fileNameCryptor(), longNameSuffix, sha1);
+		String orphanDirIdHash = dir.getParent().getFileName().toString() + dir.getFileName().toString();
+
+		Path recoveryDir = prepareRecoveryDir(pathToVault, cryptor.fileNameCryptor());
+		if (!recoveryDir.toAbsolutePath().equals(orphanedDir.toAbsolutePath())) {
```
overheadhunter, line 69: I wonder if `if (equal) { return; }` might be more readable, since this is some kind of "precondition check"...

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -108,45 +110,47 @@ public void fix(Path pathToVault, VaultConfig config, Masterkey masterkey, Crypt
...
+	CryptoPathMapper.CiphertextDirectory prepareStepParent(Path cipherRecoveryDir, FileNameCryptor cryptor, String clearStepParentDirName) throws IOException {
 		//create "step-parent" directory to move orphaned files to
 		String cipherStepParentDirName = convertClearToCiphertext(cryptor, clearStepParentDirName, Constants.RECOVERY_DIR_ID);
 		Path cipherStepParentDirFile = cipherRecoveryDir.resolve(cipherStepParentDirName + "/" + Constants.DIR_FILE_NAME);
 		final String stepParentUUID;
 		if (Files.exists(cipherStepParentDirFile, LinkOption.NOFOLLOW_LINKS)) {
 			stepParentUUID = Files.readString(cipherStepParentDirFile, StandardCharsets.UTF_8);
 		} else {
-			Files.createDirectory(cipherStepParentDirFile.getParent());
+			Files.createDirectories(cipherStepParentDirFile.getParent());
```
overheadhunter: I'd prefer `createDirectory`, since it gives us a `FileAlreadyExistException` for free. And since our stepparent has a unique name, an existing directory should not be possible (or indicates some error).

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -108,45 +110,47 @@ public void fix(Path pathToVault, VaultConfig config, Masterkey masterkey, Crypt
...
-	void adoptOrphanedResource(Adoption adoption, CryptoPathMapper.CiphertextDirectory stepParentDir, FileNameCryptor cryptor, String longNameSuffix, MessageDigest sha1) throws IOException {
-		if (adoption.oldCipherPath.toString().endsWith(Constants.DEFLATED_FILE_SUFFIX)) {
-			var newCipherName = convertClearToCiphertext(cryptor, adoption.newClearname + longNameSuffix, stepParentDir.dirId);
+	void adoptOrphanedResource(Path oldCipherPath, String newClearname, CryptoPathMapper.CiphertextDirectory stepParentDir, FileNameCryptor cryptor, String longNameSuffix, MessageDigest sha1) throws IOException {
+		if (oldCipherPath.toString().endsWith(Constants.DEFLATED_FILE_SUFFIX)) {
+			var newCipherName = convertClearToCiphertext(cryptor, newClearname + longNameSuffix, stepParentDir.dirId);
 			var deflatedName = BaseEncoding.base64Url().encode(sha1.digest(newCipherName.getBytes(StandardCharsets.UTF_8))) + Constants.DEFLATED_FILE_SUFFIX;
 			Path targetPath = stepParentDir.path.resolve(deflatedName);
-			Files.move(adoption.oldCipherPath, targetPath, StandardCopyOption.ATOMIC_MOVE);
+			Files.move(oldCipherPath, targetPath, StandardCopyOption.ATOMIC_MOVE);
```
overheadhunter: I believe ATOMIC_MOVE may thrown an exception, if the fs doesn't support it.

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -108,45 +110,47 @@ public void fix(Path pathToVault, VaultConfig config, Masterkey masterkey, Crypt
...
 			//adjust name.c9s
 			try (var fc = Files.newByteChannel(targetPath.resolve(Constants.INFLATED_FILE_NAME), StandardOpenOption.WRITE, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)) {
 				fc.write(ByteBuffer.wrap(newCipherName.getBytes(StandardCharsets.UTF_8)));
 			}
 		} else {
-			var newCipherName = convertClearToCiphertext(cryptor, adoption.newClearname, stepParentDir.dirId);
+			var newCipherName = convertClearToCiphertext(cryptor, newClearname, stepParentDir.dirId);
 			Path targetPath = stepParentDir.path.resolve(newCipherName);
-			Files.move(adoption.oldCipherPath, targetPath, StandardCopyOption.ATOMIC_MOVE);
+			Files.move(oldCipherPath, targetPath, StandardCopyOption.ATOMIC_MOVE);
```
overheadhunter: same

overheadhunter (changes requested):
great improvement on modularity and tests! just some minor open questions left for discussion

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
infeo: Consider the following edge case:
1. a fix() is incomplete
2. the step-parent becomes orphaned

Due to the incomplete fix() the (actual) orphan still exists and with `createDirectory` every subsequent fix() attempt fails, because the stepparent name is deterministically generated. In contrast, when using `createDirectories` the fix can just continue where it left.

infeo: My opinion is that in general inside the `CRYPTMATOR_RECOVERY` dir whe should be as lenient as possible wiht existing resources ar any errors. If a user stores normal data inside, its his/her fault.

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
infeo: What do you suggest? I vote for a retry without the parameter, maybe even caching the setting afterwards.

overheadhunter: I'm not entirely sure about my assumption. But if I'm not mistaken, I'd vote for just going straight for non-atomic moves. Since we move files to targets that we know for sure don't exist yet and furthermore will not access these files until the user exists health check and unlocks the vault again, there shouldn't be much need to avoid race conditions. Treat the file system as eventually consistent.

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
overheadhunter: I see, I didn't had in mind that the stepparent name is essentially the original dir id hash. I thought it would be "more unique" but I do see the benefits of continuing an incomplete fix in the same dir.

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -62,12 +62,16 @@ public String toString() {
...
+		Path orphanedDir = dataDir.resolve(this.dir);
 		String orphanDirIdHash = dir.getParent().getFileName().toString() + dir.getFileName().toString();
 
 		Path recoveryDir = prepareRecoveryDir(pathToVault, cryptor.fileNameCryptor());
-		if (!recoveryDir.toAbsolutePath().equals(orphanedDir.toAbsolutePath())) {
-			var stepParentDir = prepareStepParent(recoveryDir, cryptor.fileNameCryptor(), orphanDirIdHash);
+		if (recoveryDir.toAbsolutePath().equals(orphanedDir.toAbsolutePath())) {
+			// recovery dir was orphaned, already recovered by prepare method
+			return;
+		} else {
```
overheadhunter, line 73: I just meant to use `if`, not `if .. else ..` 😅

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
infeo: Good point about the access restriction. I'll remove the option.

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -68,28 +67,27 @@ public void fix(Path pathToVault, VaultConfig config, Masterkey masterkey, Crypt
...
-				for (Path orphanedResource : orphanedContentStream) {
-					var newClearName = switch (determineCiphertextFileType(orphanedResource)) {
+		var stepParentDir = prepareStepParent(dataDir, recoveryDir, cryptor.fileNameCryptor(), orphanDirIdHash);
+		AtomicInteger fileCounter = new AtomicInteger(1);
+		AtomicInteger dirCounter = new AtomicInteger(1);
+		AtomicInteger symlinkCounter = new AtomicInteger(1);
+		String longNameSuffix = createClearnameToBeShortened(config.getShorteningThreshold());
+		try (var orphanedContentStream = Files.newDirectoryStream(orphanedDir)) {
+			for (Path orphanedResource : orphanedContentStream) {
+				//@formatter:off
```
overheadhunter, line 80: what happens without `@formatter:off`? The switch expression looks pretty normal to me

infeo: the formatter aligns the closing bracket of the switch expression with the line start. i'd rather want to have it indented. The formatter comments will keep this in subsequent reformats.

overheadhunter (approved):

PR #108, opened 2021-07-19
Health API / Directory Check: Result EmptyDirFile

infeo:
This PR adds to the Directory Check the `EmptyDirFileResult`.

It indicates, that a dir.c9r file exists, but it is empty.

The suggested, but yet not implemented fix is to remove this dir file and its parent, such that the name is not blocked in this directory without any apparent reason for the user.

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/health/dirid/EmptyDirFile.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/dirid/EmptyDirFile.java
@@ -8,7 +8,7 @@ import java.util.Map;
 import static org.cryptomator.cryptofs.health.api.CommonDetailKeys.DIR_ID_FILE;
 
 /**
- * A critical diagnostic result of an empty directory id file (dir.c9r).
+ * A diagnostic result of an empty directory id file (dir.c9r).
  * <p>
  * Even though the empty directory ID exists, it is reserved for the root node of the crypto filesystem only.
  * Due to its nature, the root node has no corresponding dir.c9r file.
--- a/src/main/java/org/cryptomator/cryptofs/health/dirid/EmptyDirFile.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/dirid/EmptyDirFile.java
@@ -1,17 +1,21 @@
 package org.cryptomator.cryptofs.health.dirid;
 
-import org.cryptomator.cryptofs.VaultConfig;
 import org.cryptomator.cryptofs.health.api.DiagnosticResult;
-import org.cryptomator.cryptolib.api.Cryptor;
-import org.cryptomator.cryptolib.api.Masterkey;
 
-import java.io.IOException;
-import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Map;
 
 import static org.cryptomator.cryptofs.health.api.CommonDetailKeys.DIR_ID_FILE;
 
+/**
+ * A critical diagnostic result of an empty directory id file (dir.c9r).
+ * <p>
+ * Even thou the empty directory ID exists, it is reserved for the root node of the crypto filesystem only.
+ * DUe to its nature, the root node has no corresponding dir.c9r file.
+ * As a consequence, actual dir.c9r files must not be empty, otherwise it is an error in the vault structure.
+ *
+ * @see org.cryptomator.cryptofs.common.Constants#ROOT_DIR_ID
+ */
 public class EmptyDirFile implements DiagnosticResult {
 
 	final Path dirFile;
--- a/src/main/java/org/cryptomator/cryptofs/health/dirid/EmptyDirFile.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/dirid/EmptyDirFile.java
@@ -1,0 +1,45 @@
+package org.cryptomator.cryptofs.health.dirid;
+
+import org.cryptomator.cryptofs.VaultConfig;
+import org.cryptomator.cryptofs.health.api.DiagnosticResult;
+import org.cryptomator.cryptolib.api.Cryptor;
+import org.cryptomator.cryptolib.api.Masterkey;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Map;
+
+import static org.cryptomator.cryptofs.health.api.CommonDetailKeys.DIR_ID_FILE;
+
+public class EmptyDirFile implements DiagnosticResult {
+
+	final Path dirFile;
+
+	public EmptyDirFile(Path dirFile) {this.dirFile = dirFile;}
+
+	@Override
+	public Severity getSeverity() {
+		return Severity.CRITICAL;
+	}
+
+	@Override
+	public String toString() {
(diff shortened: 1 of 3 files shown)


src/test/java/org/cryptomator/cryptofs/health/dirid/DirIdCheckTest.java
```diff
@@ -130,7 +143,11 @@ private static void initDirStructure(Path root, String line) throws UncheckedIOE
 				var file = line.substring(0, sep);
 				var contents = line.substring(sep + 3);
 				Files.createDirectories(root.resolve(file).getParent());
-				Files.writeString(root.resolve(file), contents, StandardCharsets.US_ASCII, StandardOpenOption.CREATE_NEW);
+				if (!contents.equals("EMPTY")) {
```
overheadhunter, line 146: 1. please invert this if
2. not sure if `EMPTY` is a sufficiently indicating that this is a magic value. Maybe use `[EMPTY]` or something that differentiates more from "normal" contents

overheadhunter: Isn't `var contents` already an empty string, if nothing follows after `sep`?

src/main/java/org/cryptomator/cryptofs/health/dirid/EmptyDirFile.java
```diff
@@ -0,0 +1,45 @@
...
+import org.cryptomator.cryptolib.api.Masterkey;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Map;
+
+import static org.cryptomator.cryptofs.health.api.CommonDetailKeys.DIR_ID_FILE;
+
+public class EmptyDirFile implements DiagnosticResult {
```
overheadhunter, line 15: Add some javadoc explaining that empty dir ids are reserved for the root dir only.

overheadhunter (changes requested):
90% good, just not sure about the magic value used during test setup

src/test/java/org/cryptomator/cryptofs/health/dirid/DirIdCheckTest.java
infeo: Nope, because seperator is " = " (note the leading and trailing whitespaces). A text block line is ended with a non-whitespace character, thus writing nothing would lead to a line ending with " =" and the splitting would fail.

There are workarounds for this, e.g. https://awesomely-java.blogspot.com/2021/03/techniques-for-including-trailing.html.

I would go with [EMPTY]. It's our code, we don't need the best approach, only a suffiecently future-proof and easy-to-understand one.

src/main/java/org/cryptomator/cryptofs/health/dirid/EmptyDirFile.java
```diff
@@ -0,0 +1,49 @@
...
+import java.nio.file.Path;
+import java.util.Map;
+
+import static org.cryptomator.cryptofs.health.api.CommonDetailKeys.DIR_ID_FILE;
+
+/**
+ * A critical diagnostic result of an empty directory id file (dir.c9r).
+ * <p>
+ * Even thou the empty directory ID exists, it is reserved for the root node of the crypto filesystem only.
+ * DUe to its nature, the root node has no corresponding dir.c9r file.
```
overheadhunter, line 14:
```suggestion
 * Even though the empty directory ID exists, it is reserved for the root node of the crypto filesystem only.
 * Due to its nature, the root node has no corresponding dir.c9r file.
```

Please no slang outside of Slack

src/main/java/org/cryptomator/cryptofs/health/dirid/EmptyDirFile.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.cryptomator.cryptofs.health.api.DiagnosticResult;
+
+import java.nio.file.Path;
+import java.util.Map;
+
+import static org.cryptomator.cryptofs.health.api.CommonDetailKeys.DIR_ID_FILE;
+
+/**
+ * A critical diagnostic result of an empty directory id file (dir.c9r).
```
overheadhunter, line 11: It is only "critical" (i.e. no fix available), because there is no fix implemented _yet_, right?

src/test/java/org/cryptomator/cryptofs/health/dirid/DirIdCheckTest.java
```diff
@@ -130,7 +143,11 @@ private static void initDirStructure(Path root, String line) throws UncheckedIOE
 				var file = line.substring(0, sep);
 				var contents = line.substring(sep + 3);
 				Files.createDirectories(root.resolve(file).getParent());
-				Files.writeString(root.resolve(file), contents, StandardCharsets.US_ASCII, StandardOpenOption.CREATE_NEW);
+				if (!contents.equals("[EMPTY]")) {
```
overheadhunter, line 146: Still: Please invert the if/else branches. Else should be the "default" case

overheadhunter (changes requested):

src/main/java/org/cryptomator/cryptofs/health/dirid/EmptyDirFile.java
infeo: Yes. I planned to propose this change to all DiagnosticResult classes as long as no fix is implemented for those.

overheadhunter: Fair enough, but I'm afraid the JavaDoc won't be adjusted too often and shouldn't reflect the work-in-progress but rather the intended role of this class.

src/test/java/org/cryptomator/cryptofs/health/dirid/DirIdCheckTest.java
infeo: Fixed in 0df4948

src/main/java/org/cryptomator/cryptofs/health/dirid/EmptyDirFile.java
infeo: Fixed in 0df4948

src/test/java/org/cryptomator/cryptofs/health/dirid/DirIdCheckTest.java
overheadhunter: Now you're opening, closing, reopening, writing, closing... No more atomic checks. I don't understand what would be so wrong about a simple `if (contents.equals("[EMPTY]"))`. But since this is just test code, I give up.

overheadhunter (approved):

src/test/java/org/cryptomator/cryptofs/health/dirid/DirIdCheckTest.java
infeo: Whoopsie 😅 
grafik

I read "revert" rather than "invert" and wondered what should be reverted... i'll fix this asap.

overheadhunter (approved):
<3

PR #109, opened 2021-07-27
Feature/type check

infeo:
This PR adds a new health check to the health service: CiphertextFileTypeCheck.

The checks looks at all c9r and c9s directories and tries to determine their cleartext type by checking for existence of "signature" files, i.e. `dir.c9r`, `symlink.c9r` and `contents.c9r`.

It produces three results: KnownType, AmbiguousType and UnknownType.

The result resolution is as follows:
1. Exists only one, _valid_ signature file, a KnownType is returned
2. Exist at least two _valid_ signature files, an AmbiguousType is returned
3. Is no valid signatre file present, an UnknownType is returned.

Valid sig files are for c9r dirs `dir.c9r` and `symlink.c9r`. For c9s its the same with additionally `contents.c9r`.


src/main/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheck.java
```diff
@@ -0,0 +1,160 @@
...
+			boolean isDir = attrs.isDirectory();
+
+			if (isDir && dirName.endsWith(Constants.CRYPTOMATOR_FILE_SUFFIX)) {
+				return DirType.C9R_DIR;
+			} else if (isDir && dirName.endsWith(Constants.DEFLATED_FILE_SUFFIX)) {
+				return DirType.C9S_DIR;
+			} else {
+				return DirType.UNRELATED;
+			}
+		}
```
overheadhunter, line 86: if you only check for `.c9r` and `.c9s` dirs (not _files_), you can simplify this by using `preVisitDirectory(...)` instead

src/main/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheck.java
```diff
@@ -0,0 +1,160 @@
...
+
+		void checkCiphertextTypeC9s(Path dir) {
+			boolean isDir = containsDirFile(dir);
+			boolean isSymlink = containsSymlinkFile(dir);
+			boolean isFile = containsContentsFile(dir);
+			var types = ciphertextFileTypesFrom(isDir, isSymlink, isFile);
+
+			if ((isDir && isFile && isSymlink) ^ isSymlink ^ isFile ^ isDir) {
```
overheadhunter, line 108: `isDir && isFile && isSymlink` → known type??

src/main/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheck.java
```diff
@@ -0,0 +1,160 @@
...
+			boolean isSymlink = containsSymlinkFile(dir);
+			var types = ciphertextFileTypesFrom(isDir, isSymlink, false);
+
+			if (isDir ^ isSymlink) {
+				resultCollector.accept(new KnownType(dir, types.iterator().next()));
+			} else if (isDir || isSymlink) {
+				resultCollector.accept(new AmbiguousType(dir, types));
+			} else {
+				resultCollector.accept(new UnknownType(dir));
+			}
```
overheadhunter, line 99: second branch is essentially `if not a xor b and a or b`, which can also be expressed as `if a and b` which is easier to read imho

src/main/java/org/cryptomator/cryptofs/health/type/KnownType.java
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	@Override
+	public Severity getSeverity() {
+		return Severity.GOOD;
+	}
+
+	@Override
+	public String toString() {
+		return String.format("Encrypted dir %s with determined type %s.", cipherDir, type);
```
overheadhunter, line 33: "Encrypted dir" suggests that this represents a cleartext dir.

```suggestion
		return String.format("Node %s with determined type %s.", cipherDir, type);
```

src/main/java/org/cryptomator/cryptofs/health/type/UnknownType.java
```diff
@@ -0,0 +1,45 @@
...
+	TODO: remove directory? might cause data loss (for shortened files)
+	@Override
+	public void fix(Path pathToVault, VaultConfig config, Masterkey masterkey, Cryptor cryptor) throws IOException {
+		DiagnosticResult.super.fix(pathToVault, config, masterkey, cryptor);
+	}
+	 */
+
+	@Override
+	public String toString() {
+		return String.format("Encrypted dir %s of unknown type.", cipherDir);
```
overheadhunter, line 38:
```suggestion
		return String.format("Node %s of unknown type.", cipherDir);
```

src/main/java/org/cryptomator/cryptofs/health/type/AmbiguousType.java
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	@Override
+	public Severity getSeverity() {
+		return Severity.CRITICAL;
+	}
+
+	@Override
+	public String toString() {
+		return String.format("Encrypted dir %s of ambiguous type. Possible types are: %s", cipherDir, possibleTypes);
```
overheadhunter, line 34:
```suggestion
		return String.format("Node %s of ambiguous type. Possible types are: %s", cipherDir, possibleTypes);
```

src/main/java/org/cryptomator/cryptofs/health/type/AmbiguousType.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A ciphertext dir ending with c9r or c9s but does contain more than one valid signature file.
+ * <p>
+ * Valid signature files for c9r are {@value org.cryptomator.cryptofs.common.Constants#DIR_FILE_NAME} and {@value org.cryptomator.cryptofs.common.Constants#SYMLINK_FILE_NAME}.
+ * Valid signature files for c9s are the ones for c9r and {@value org.cryptomator.cryptofs.common.Constants#CONTENTS_FILE_NAME}.
+ */
+public class AmbiguousType implements DiagnosticResult {
```
overheadhunter, line 17: to draft a fix: Create sibling nodes with "_2", "_3", for each possible type and move one type-specific files to them

src/test/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheckTest.java
```diff
@@ -0,0 +1,227 @@
...
+				MatcherAssert.assertThat(resultCaptor.getValue(), Matchers.instanceOf(UnknownType.class));
+			}
+
+			@ParameterizedTest
+			@DisplayName("contents.c9r is ignored for c9r dir")
+			@MethodSource("provideAllContentsC9rCombos")
+			public void testContentsC9rIsIgnored(List<String> sigFiles, Class resultClass) throws IOException {
+				for (var sigFile : sigFiles) {
+					Files.createFile(c9rDir.resolve(sigFile));
+				}
```
overheadhunter, line 144: `sigFiles`?

src/main/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheck.java
overheadhunter: actually it is impossible for all three to be true, when xored with each of these values individually

src/main/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheck.java
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		void checkCiphertextTypeC9s(Path dir) {
+			boolean isDir = containsDirFile(dir);
+			boolean isSymlink = containsSymlinkFile(dir);
+			boolean isFile = containsContentsFile(dir);
```
overheadhunter, line 105: I wonder if a bit field might be easier to operate on 🤔

overheadhunter (changes requested):

src/main/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheck.java
```diff
@@ -0,0 +1,160 @@
...
+		public DirVisitor(Consumer<DiagnosticResult> resultCollector) {this.resultCollector = resultCollector;}
+
+		@Override
+		public FileVisitResult visitFile(Path dir, BasicFileAttributes attrs) {
+			switch (determineFileType(dir, attrs)) {
+				case C9R_DIR -> checkCiphertextTypeC9r(dir);
+				case C9S_DIR -> checkCiphertextTypeC9s(dir);
+				case UNRELATED -> {}
+			}
+			;
```
overheadhunter, line 71: remove line

src/main/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheck.java
overheadhunter: In fact, the easiest approach would probably be to just count the number of items in `types`.

src/main/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheck.java
overheadhunter: In fact, the easiest approach would probably be to just count the number of items in `types`.

src/test/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheckTest.java
infeo: sigFiles = signatureFiles

renamed in f3c94769af0aedb0c738d37f33cb0c2e2a608d13

src/main/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheck.java
infeo: good idea! I decided to use an octal system, similar to the linux file permissions.

Changed in eba145b

src/main/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheck.java
infeo: > just count the number of items in `types`:

Indeed a very good idea, thanks for the code snippet! Fixed in bd78457

src/main/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheck.java
infeo: Fixed in bd78457

src/main/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheck.java
infeo: That is true, (before f138d82 it was the case)  but then i need to increase `MAX_DEPTH` to 4 and should return `SKIP_SUBTREE` if a c9r/c9s dir is detected. Hence, the "complexity" will be just shifted elsewhere. Also, with an increased MAX_DEPTH we could visit more _unrelated_ files/directories.

Also performant wise there is nearly no gain: Using only the `preVisitDirectory(...)` method, the `visitFile(...)` has less no branching (due to SimpleFileVisitor) and under the assumption there are always more files than directories to visit, it should be faster. But the time-consuming task is reading the attributes and we do not skip any files or directories,

src/test/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheckTest.java
overheadhunter: jup I saw signatureFiles elsewhere, however I am wondering what this has to do with signatures?

infeo: When I used the term i had "signature file" as in "signature move" in mind: An identifying file.

overheadhunter: Oh ok, got confused due to the cryptography context. Let's stick to "type" used elsewhere, i.e. `typeFiles`

src/main/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheck.java
overheadhunter: ok. btw I didn't know that `visitFile` is invoked for directories as well

infeo: > ok. btw I didn't know that `visitFile` is invoked for directories as well

Well yes and no. Directories are only visited at maxDepth, see [Files.walkFileTree](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/nio/file/Files.html#walkFileTree(java.nio.file.Path,java.util.Set,int,java.nio.file.FileVisitor)), third last paragraph.

src/test/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheckTest.java
infeo: Fixed in 765c011

src/main/java/org/cryptomator/cryptofs/health/type/AmbiguousType.java
infeo: Feasible, but it requires to build up the decrypted dir structure up to directory containing this ambiguous type because we need the dirId to get/create a cleartext filename. In the worst case the fix fails because we are inside an orphaned directory, for which no dirId exists anymore.

Actually, i think, for several checks knowing this structure would be benefical (e.g. providing decrypted names in `DiagnosticResult::details`), and hence suggest to build it in the first check requiring it and then caching it for later checks (done with lazy dagger injection?). Of course this would need a not too small amount of code to write and test and is out of scope for this PR.

src/main/java/org/cryptomator/cryptofs/health/type/CiphertextFileTypeCheck.java
overheadhunter: actually this suggestion got superseeded by bd78457, as there is no longer the need to do "ifs" on the results (as in `bitmask & FILE == FILE`)

I wrote this suggestion before the "count the number of detected types" thing. Sorry!

overheadhunter:
I figured the `containsDirFile(...)`, `containsSymlinkFile(...)` etc are only used for a single purpose, so I inlined them in c005cb5 and got rid of some code (+22 -46).

Before approving, I'd still like to discuss, if we really should skip checking for misplaced `unshortened.c9r/contents.c9r` files (in other words: get rid of `checkForContentsC9r` flag). In theory this case shouldn't exist. But we're doing a health check, after all...

overheadhunter (approved):

PR #110, opened 2021-08-09
Feature/shortened names check

infeo:
This PR adds a new check to the health check API: ShortenedNamesCheck

For all directories ending with `c9s` it checks, if the file `name.c9s` exists in its hashed & encoded content matches the directory name.

It produces 4 results:
* Valid shortened resource
* Obese result, if the size of `name.c9s` is bigger than a cryptofs internal limit
* Missing result, if `name.c9s` is not a regular file or it does not exist
* Mismatch result, if the directory name does not match the hashed & encoded content

For the future this check can be refinded by providing the directory id of the encrypted dir containing the shortened resource, but for this PR it is out of scope and only noted via inline comments.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,8 +143,10 @@
 				<version>3.0.0-M5</version>
 				<configuration>
 					<!-- Allow reflection for Mockito, so it can properly mock non-public classes -->
-					<argLine>--add-opens=org.cryptomator.cryptofs/org.cryptomator.cryptofs.health.dirid=ALL-UNNAMED --add-opens=org.cryptomator.cryptofs/org.cryptomator.cryptofs.health.type=ALL-UNNAMED</argLine>
-				</configuration>
+					<argLine>--add-opens=org.cryptomator.cryptofs/org.cryptomator.cryptofs.health.dirid=ALL-UNNAMED
+						--add-opens=org.cryptomator.cryptofs/org.cryptomator.cryptofs.health.type=ALL-UNNAMED
+						--add-opens=org.cryptomator.cryptofs/org.cryptomator.cryptofs.health.shortened=ALL-UNNAMED</argLine>
+			</configuration>
 			</plugin>
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
--- a/src/main/java/org/cryptomator/cryptofs/health/shortened/MissingLongName.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/shortened/MissingLongName.java
@@ -1,0 +1,43 @@
+package org.cryptomator.cryptofs.health.shortened;
+
+import org.cryptomator.cryptofs.VaultConfig;
+import org.cryptomator.cryptofs.health.api.DiagnosticResult;
+import org.cryptomator.cryptolib.api.Cryptor;
+import org.cryptomator.cryptolib.api.Masterkey;
+
+import java.io.IOException;
+import java.nio.file.Path;
+
+/**
+ * TODO: doc doc doc
+ * 			- the duckumentation duck
+ *		   __
+ *	   ___( o)>
+ *	   \ <_. )
+ *		`---'   hjw
+ */
+public class MissingLongName implements DiagnosticResult {
+
+	final Path c9sDir;
+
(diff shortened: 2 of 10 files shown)


src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
```diff
@@ -0,0 +1,131 @@
...
+			}
+			if (!attrs.isRegularFile()) {
+				resultCollector.accept(new MissingLongName(dir));
+				return;
+			} else if (attrs.size() > LongFileNameProvider.MAX_FILENAME_BUFFER_SIZE) {
+				resultCollector.accept(new ObeseNameFile(nameFile, attrs.size()));
+				return;
+			}
+
+			var longName = readLongName(nameFile);
```
overheadhunter, line 102: Since we can now facilitate new APIs, let's get rid of `readLongName (...)` \*

```suggestion
			var longName = Files.readString(nameFile, UTF_8);
```

\* just here, not in LongFileNameProvider - that can be done as a separate commit directly on develop eventually

pom.xml
```diff
@@ -143,8 +143,10 @@
 				<version>3.0.0-M5</version>
 				<configuration>
 					<!-- Allow reflection for Mockito, so it can properly mock non-public classes -->
-					<argLine>--add-opens=org.cryptomator.cryptofs/org.cryptomator.cryptofs.health.dirid=ALL-UNNAMED --add-opens=org.cryptomator.cryptofs/org.cryptomator.cryptofs.health.type=ALL-UNNAMED</argLine>
-				</configuration>
+					<argLine>--add-opens=org.cryptomator.cryptofs/org.cryptomator.cryptofs.health.dirid=ALL-UNNAMED
+						--add-opens=org.cryptomator.cryptofs/org.cryptomator.cryptofs.health.type=ALL-UNNAMED
+						--add-opens=org.cryptomator.cryptofs/org.cryptomator.cryptofs.health.shortened=ALL-UNNAMED</argLine>
```
overheadhunter, line 148: really required? what non-public class is mocked in the tests?

overheadhunter (changes requested):
Looks good, except for the various comments revolving around the possible future improvements when a DirID may be present. _This_ feature is feature-complete, so remove those comments. It'd be a different PR to deal with them.

src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
infeo: Just a remark about the future replacement in LongFileNameProvider:  `Files.readString(...)` has no size restrictions of the file to read. This should be kept in mind. Here it does not matter, because the size is checked beforehand.

infeo:
>This feature is feature-complete, so remove those comments

Should I only remove the "TODO" marker or the whole comment?

pom.xml
overheadhunter: 👍

overheadhunter:
> Should I only remove the "TODO" marker or the whole comment?

The whole comment. Treat the current implementation as final, document future changes in an issue.

overheadhunter (approved):
Still contains an unspecific `// TODO` but otherwise ready to merge

src/main/java/org/cryptomator/cryptofs/health/shortened/MissingLongName.java
```diff
@@ -0,0 +1,33 @@
...
+	public MissingLongName(Path c9sDir) {this.c9sDir = c9sDir;}
+
+	@Override
+	public Severity getSeverity() {
+		return Severity.CRITICAL;
+	}
+
+	@Override
+	public String toString() {
+		return String.format("Shortened resource %s either misses %s or the file has invalid content.", c9sDir, Constants.INFLATED_FILE_NAME); //TODO
```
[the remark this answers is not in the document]
infeo: Fixed in https://github.com/cryptomator/cryptofs/pull/110/commits/18e4959d01db6c36b0eb2b3a193c1842a4a8ab69

PR #117, opened 2021-10-31
Generalize backup functionality (in preperation for: https://github.com/cryptomator/cryptomator/issues/1825)

barthr:


src/main/java/org/cryptomator/cryptofs/common/BackupHelper.java
```diff
@@ -40,22 +40,23 @@ public static String generateFileIdSuffix(byte[] fileBytes) {
...
-	 * @throws IOException If the masterkey cannot be read.
+	 * @param path The file to back up
+	 * @throws IOException If the path cannot be read.
 	 */
-	public static Path attemptMasterKeyBackup(Path masterKeyPath) throws IOException {
-		byte[] keyFileContents = Files.readAllBytes(masterKeyPath);
-		String backupFileName = masterKeyPath.getFileName().toString() + generateFileIdSuffix(keyFileContents) + Constants.MASTERKEY_BACKUP_SUFFIX;
-		Path backupFilePath = masterKeyPath.resolveSibling(backupFileName);
+	public static Path attemptBackup(Path path) throws IOException {
+		byte[] keyFileContents = Files.readAllBytes(path);
```
overheadhunter, line 50: rename this to `fileContents`

src/main/java/org/cryptomator/cryptofs/common/BackupHelper.java
```diff
@@ -64,16 +65,17 @@ private static void assertExistingBackupMatchesContent(Path backupFilePath, Byte
 		if (Files.exists(backupFilePath)) {
 			// TODO replace by Files.mismatch() when using JDK > 12
```
overheadhunter, line 66: While you're at it: Since we're now on JDK 17, we can make use of `Files.mismatch()`. In other words, instead of comparing file to the contents of a bytebuffer, we can refactor this method to `assertSameContent(Path backupFile, Path originalFile)` and simplify it.

overheadhunter: something like `if (Files.mismatch(...) != -1) { log warning }`

overheadhunter (changes requested):
Sorry for taking so long to review this. Looks promising, I just found one more variable to rename and a chance to simplify code.

src/main/java/org/cryptomator/cryptofs/common/BackupHelper.java
barthr: Sounds good! will do

overheadhunter (approved):

PR #122, opened 2021-12-15
Feature: Do not add nullbytes when persisting Long Filename

### LINKED ISSUE #121: Long filesnames are persisted with NULL Bytes at the end
In CryptoFS, the name of a resource (file, dir, symlink) is shortend and the name is stored in an additional file (`name.c9s`), if the name exceeds a certain length. In version 2.3.0. This mechanism is broken in cryptofs 2.3.0.

When a file with a long name (aka it needs to be shortend) is created, in `name.c9s` NULL-Bytes might be appended at the end of the encrypted long name.

The reason for this behaviour is f8079199c313942cf83efe134c16aca629a9336a. `ByteBuffer.array()` returns the _whole_ array of the byte buffer, not just the "non-empty" part of it.

infeo:
This closes #121.

It adds two additional results to the shortend health check:
* `NotDecodableLongName` if the long filename ist not base64url encoded
* `TrailingNullBytes` if the stored string in name.c9s is affected by issue #121

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
@@ -24,6 +24,7 @@ import java.nio.file.attribute.BasicFileAttributeView;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Set;
 import java.util.function.Consumer;
+import java.util.regex.Pattern;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.cryptomator.cryptofs.common.Constants.DEFLATED_FILE_SUFFIX;
@@ -37,6 +38,7 @@ public class ShortenedNamesCheck implements HealthCheck {
 	private static final Logger LOG = LoggerFactory.getLogger(ShortenedNamesCheck.class);
 	private static final int MAX_TRAVERSAL_DEPTH = 3;
 	private static final BaseEncoding BASE64URL = BaseEncoding.base64Url();
+	private static final Pattern VALID_EXCEPT_NULLBYTE_END = Pattern.compile("[^\0]+\0+");
 
 	@Override
 	public String name() {
@@ -95,6 +97,13 @@ public class ShortenedNamesCheck implements HealthCheck {
 			}
 
 			var longName = Files.readString(nameFile, UTF_8);
+
+			//checking for https://github.com/cryptomator/cryptofs/issues/121
+			if(VALID_EXCEPT_NULLBYTE_END.matcher(longName).matches()) {
+				resultCollector.accept(new TrailingNullBytesInNameFile(nameFile, longName));
+				return;
+			}
+
 			var expectedShortName = deflate(longName);
 			if (!dir.getFileName().toString().equals(expectedShortName)) {
 				resultCollector.accept(new LongShortNamesMismatch(dir, expectedShortName));
--- a/src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
@@ -38,7 +38,7 @@ public class ShortenedNamesCheck implements HealthCheck {
 	private static final Logger LOG = LoggerFactory.getLogger(ShortenedNamesCheck.class);
 	private static final int MAX_TRAVERSAL_DEPTH = 3;
 	private static final BaseEncoding BASE64URL = BaseEncoding.base64Url();
-	private static final Pattern VALID_EXCEPT_NULLBYTE_END = Pattern.compile("[^\0]+\0+");
+	private static final Pattern NULL_BYTES = Pattern.compile("\0+");
 
 	@Override
 	public String name() {
@@ -98,8 +98,11 @@ public class ShortenedNamesCheck implements HealthCheck {
 
 			var longName = Files.readString(nameFile, UTF_8);
 
-			//checking for https://github.com/cryptomator/cryptofs/issues/121
-			if(VALID_EXCEPT_NULLBYTE_END.matcher(longName).matches()) {
+			var result = isValidEncodedName(longName);
+			if (result == NameEncoding.INVALID) {
+				resultCollector.accept(new NotDecodableLongName(nameFile, longName));
+				return;
+			} else if (result == NameEncoding.BUG121) {
 				resultCollector.accept(new TrailingNullBytesInNameFile(nameFile, longName));
 				return;
 			}
@@ -110,6 +113,36 @@ public class ShortenedNamesCheck implements HealthCheck {
 			} else {
 				resultCollector.accept(new ValidShortenedFile(dir));
 			}
+		}
+
+
+		/**
+		 * Checks if the string stored inside the name file is base64url encoded
+		 *
+		 * <em>visible for testing</em>
+		 *
+		 * @return {@link NameEncoding} indicating if it is valid, invalid or affected by https://github.com/cryptomator/cryptofs/issues/121
+		 */
+		NameEncoding isValidEncodedName(String toAnalyse) {
+			int firstNullByte = toAnalyse.indexOf('\0');
+			var beforeFirstNull = firstNullByte == -1 ? toAnalyse : toAnalyse.substring(0, firstNullByte);
+			var afterFirstNull = firstNullByte == -1 ? "" : toAnalyse.substring(firstNullByte);
+
+			boolean isFirstBase64 = BASE64URL.canDecode(beforeFirstNull);
+
+			if (isFirstBase64 && afterFirstNull.isEmpty()) {
+				return NameEncoding.VALID;
+			} else if (isFirstBase64 && NULL_BYTES.matcher(afterFirstNull).matches()) {
+				return NameEncoding.BUG121;
+			} else {
+				return NameEncoding.INVALID;
+			}
+		}
+
+		enum NameEncoding {
+			VALID,
+			INVALID,
+			BUG121;
 		}
 
 		//visible for testing
--- a/src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
@@ -91,19 +91,19 @@ public class ShortenedNamesCheck implements HealthCheck {
 			if (!attrs.isRegularFile()) {
 				resultCollector.accept(new MissingLongName(dir));
 				return;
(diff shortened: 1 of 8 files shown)


src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
```diff
@@ -103,6 +115,36 @@ void checkShortenedName(Path dir) throws IOException {
...
+				return NameEncoding.BUG121;
+			} else {
+				return NameEncoding.INVALID;
+			}
+		}
+
+		enum NameEncoding {
+			VALID,
+			INVALID,
+			BUG121;
```
overheadhunter, line 145: I'd prefer a name that doesn't require you to know the issue description. A link to the github issue in a comment would be nice, though.

Suggestion:

```suggestion
			NULL_PADDED; // contains trailing 0x00 bytes, see https://github.com/cryptomator/cryptofs/issues/121
```

src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
```diff
@@ -103,6 +115,36 @@ void checkShortenedName(Path dir) throws IOException {
...
+			boolean isFirstBase64 = BASE64URL.canDecode(beforeFirstNull);
+
+			if (isFirstBase64 && afterFirstNull.isEmpty()) {
+				return NameEncoding.VALID;
+			} else if (isFirstBase64 && NULL_BYTES.matcher(afterFirstNull).matches()) {
+				return NameEncoding.BUG121;
+			} else {
+				return NameEncoding.INVALID;
+			}
+		}
```
overheadhunter, line 140: I think cutting the string in halves at the first occurrence of a null char is overly complicated.

```suggestion
		NameEncoding isValidEncodedName(String toAnalyse) {
			var trimmed = toAnalyse.trim();
			return !toAnalyse.equals(trimmed) //
				? NameEncoding.BUG121 // string padded with chars < 0x20
				: BASE64URL.canDecode(toAnalyse) ? NameEncoding.VALID : NameEncoding.INVALID;
		}
```

src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
```diff
@@ -103,6 +115,36 @@ void checkShortenedName(Path dir) throws IOException {
...
 
+
+		/**
+		 * Checks if the string stored inside the name file is base64url encoded
+		 *
+		 * <em>visible for testing</em>
+		 *
+		 * @return {@link NameEncoding} indicating if it is valid, invalid or affected by https://github.com/cryptomator/cryptofs/issues/121
+		 */
+		NameEncoding isValidEncodedName(String toAnalyse) {
```
overheadhunter, line 126: Since this is no longer a question of base64 encoding or not, I'd strip the word "encoding" from this method (and return type). Rather use something like `ContentRating validateC9sFileContent(String content)`.

src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
```diff
@@ -103,6 +115,36 @@ void checkShortenedName(Path dir) throws IOException {
...
+			int firstNullByte = toAnalyse.indexOf('\0');
+			var beforeFirstNull = firstNullByte == -1 ? toAnalyse : toAnalyse.substring(0, firstNullByte);
+			var afterFirstNull = firstNullByte == -1 ? "" : toAnalyse.substring(firstNullByte);
+
+			boolean isFirstBase64 = BASE64URL.canDecode(beforeFirstNull);
```
overheadhunter, line 131: doesn't the "long file name" contain a file extension? iirc, it does. and `.` is not a valid base64url char.

src/main/java/org/cryptomator/cryptofs/health/shortened/TrailingNullBytesInNameFile.java
```diff
@@ -0,0 +1,44 @@
...
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.Map;
+
+/**
+ * Result and fix for bug https://github.com/cryptomator/cryptofs/issues/121
+ */
+public class TrailingNullBytesInNameFile implements DiagnosticResult {
```
overheadhunter, line 19: Instead of just null, we might want to report (and fix) _any_ non-printable/whitespace char (ASCII < 0x20). While we don't know of a bug that adds such chars, it is certainly useful if such content is added by unknown causes and requires exactly the same fix.

overheadhunter (changes requested):
I believe we can broaden this fix for further unprintable chars. And simplify the check. And rename some methods. 😉

src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
infeo: My goal was to be as precise as possible: Only return `NameEncoding.BUG121`, if the string is base64url encoded and has trailing null bytes.

But we can widen the scope of the check a little, we should be only careful in doing so and make as less assumptions as possible. For the suggestion the assumption is not only "affected by the bug" but also "it might happen, that chars < 0x20 are added before and after the targeted string".

infeo: In general: We should keep the check, the result and its fix as narrow scoped as possible, to not fix an error in the wrong way.

There can be a lot of things broken in the vault structure, but in my opinion better stick to only the things with known causes or unambigous repairs. In the suggestion, the cause is unknown _and_ the repair is ambigous, because the string between the padding can be still invalid base64url.

src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
infeo: Fixed in db83548

src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
infeo: Fixed in db83548

src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
infeo: Considered for db83548 , I renamed it to `TRAILING_BYTES`

src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
```diff
@@ -37,6 +38,7 @@
 	private static final Logger LOG = LoggerFactory.getLogger(ShortenedNamesCheck.class);
 	private static final int MAX_TRAVERSAL_DEPTH = 3;
 	private static final BaseEncoding BASE64URL = BaseEncoding.base64Url();
+	private static final Pattern NULL_BYTES = Pattern.compile("\0+");
```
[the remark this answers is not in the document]
infeo: Removed in db83548

src/main/java/org/cryptomator/cryptofs/health/shortened/TrailingNullBytesInNameFile.java
infeo: Considered for db83548. All trailing bytes after `.c9r` are removed.

src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
infeo: After a short discussion we agreed to the following changes (as implemented in db83548 ): Make a basic syntax check by searching for `.c9r`in the long name and check for trailing bytes after and that the suffix is base64url encoded.

src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
```diff
@@ -103,6 +115,38 @@ void checkShortenedName(Path dir) throws IOException {
...
+			}
+
+			var encryptedFileName = toAnalyse.substring(0, posObligatoryC9rString);
+			if(!BASE64URL.canDecode(encryptedFileName)) {
+				return SyntaxResult.INVALID;
+			}
+
+			if (toAnalyse.substring(posObligatoryC9rString).length() > Constants.CRYPTOMATOR_FILE_SUFFIX.length() ) {
+				return SyntaxResult.TRAILING_BYTES;
+			}
```
overheadhunter, line 139: just a minor optimization to avoid creating substrings

```suggestion
			if (toAnalyse.length() > posObligatoryC9rString +  Constants.CRYPTOMATOR_FILE_SUFFIX.length()) {
				return SyntaxResult.TRAILING_BYTES;
			}

			var encryptedFileName = toAnalyse.substring(0, posObligatoryC9rString);
			if(!BASE64URL.canDecode(encryptedFileName)) {
				return SyntaxResult.INVALID;
			}
```

overheadhunter (approved):
Just a suggestion, but I'm fine to merge it as is.

src/main/java/org/cryptomator/cryptofs/health/shortened/ShortenedNamesCheck.java
infeo: The optimization is actually a change, which changes the outcome: Before, you always check the if the first part is base64url encoded and if it is not always return `INVALID` and only if it is valid, it might be `TRAILING_BYTES` result. 
If applying the change, we always check if there are trailing bytes. And only if there are none, it might return `INVALID`.

I decided against this change, since i see no benefit in fixing trailing bytes if the result would still be broken from _this checks perspective_.

overheadhunter: even without reordering, note the unnecessary substring, though:

```diff
-			if (toAnalyse.substring(posObligatoryC9rString).length() > Constants.CRYPTOMATOR_FILE_SUFFIX.length() ) {
+			if (toAnalyse.length() > posObligatoryC9rString +  Constants.CRYPTOMATOR_FILE_SUFFIX.length()) {
```

infeo: the next time 🙂

infeo: done in fe23ff483c26484c4e6a456b7d6797596f689b02

PR #125, opened 2022-01-26
Use BufferPool for cleartext as well as ciphertext chunks

overheadhunter:
Previously, each chunk of data allocated a new 32kB ByteBuffer, which was in most cases very short-lived. This lead to both, massive allocations as well as garbage collection activity.

To reduce both memory as well as CPU pressure, this introduces a filesystem-scoped pooling mechanism. ByteBuffers _should_ be returned to the pool but it will do no harm not to do so.

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java
+++ b/src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java
@@ -104,9 +104,9 @@ public class CleartextFileChannel extends AbstractFileChannel {
 			long pos = position + read;
 			long chunkIndex = pos / payloadSize; // floor by int-truncation
 			int offsetInChunk = (int) (pos % payloadSize); // known to fit in int, because payloadSize is int
-			int len = min(dst.remaining(), payloadSize - offsetInChunk); // known to fit in int, because second argument is int
-			final ChunkData chunkData = chunkCache.get(chunkIndex);
-			chunkData.copyDataStartingAt(offsetInChunk).to(dst);
+			ByteBuffer data = chunkCache.get(chunkIndex).data().duplicate().position(offsetInChunk);
+			int len = min(dst.remaining(), data.remaining()); // known to fit in int, because second argument is int
+			dst.put(data.limit(data.position() + len));
 			read += len;
 		}
 		dst.limit(origLimit);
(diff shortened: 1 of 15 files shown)


infeo (changes requested):
Amazing work you did here 👍 

@overheadhunter Did you considered the `Chunk` class implementing `Autoclosable`? In the close method, it could automatically call `bufferpool.recycle()`.

* Chunk implementing Autoclosable

src/main/java/org/cryptomator/cryptofs/fh/ChunkLoader.java
```diff
@@ -15,38 +15,36 @@
...
+		ByteBuffer ciphertextBuf = bufferPool.getCiphertextBuffer();
+		ByteBuffer cleartextBuf = bufferPool.getCleartextBuffer();
+		try {
+			int read = ciphertext.read(ciphertextBuf, ciphertextPos);
+			if (read == -1) {
+				cleartextBuf.limit(0);
 			} else {
-				cleartextBufWhichCanHoldFullChunk = cleartextBuf;
+				ciphertextBuf.flip();
+				cryptor.fileContentCryptor().decryptChunk(ciphertextBuf, cleartextBuf, chunkIndex, headerHolder.get(), true);
```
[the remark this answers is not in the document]
overheadhunter: Nah, Buffer suggests a Java type (of any size), while Chunk is a term we use in our scheme docs and is specified to be [of a certain size](https://github.com/cryptomator/cryptolib/blob/ccfe34dc0052f906ad9c840c73f715e2ae6d4837/src/main/java/org/cryptomator/cryptolib/api/FileContentCryptor.java#L22-L30).

overheadhunter:
> Did you considered the `Chunk` class implementing `Autoclosable`?

The problem is that Chunk is passed around between classes a lot, i.e. try-with-resource can't be used. It is even cached, so analysis tools have a hard time tracking the lifecycle and will report false positive leaks.

src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java
```diff
@@ -148,17 +151,21 @@ private long writeLockedInternal(ByteSource src, long position) throws IOExcepti
 			assert len <= cleartextChunkSize;
 			if (offsetInChunk == 0 && len == cleartextChunkSize) {
 				// complete chunk, no need to load and decrypt from file
-				ChunkData chunkData = ChunkData.emptyWithSize(cleartextChunkSize);
-				chunkData.copyData().from(src);
-				chunkCache.set(chunkIndex, chunkData);
+				ByteBuffer cleartextChunk = bufferPool.getCleartextBuffer();
```
infeo, line 154: should be named `cleartextBuffer`

infeo (changes requested):
Just one naming thing

src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java
overheadhunter: But it _is_ a chunk. I prefer to name variables according to semantics rather than to types.

infeo: Correct, but from a technical perspective a chunk is an object derived from the `Chunk` class wrapping a buffer. Here those two definitions intersect, and to avoid confusion I would tend to the technical description. I mean, we are wrapping three lines later a "chunk" with "Chunk". And at creation, the "chunk" is not a "chunk" but just a buffer (the method even says it). Additionally, the naming scheme is also used in other places.

As a compromise, how about renaming it to `cleartextChunkData` ?

infeo (approved):

PR #127, opened 2022-04-05
Use dirid file in orphan dir fix

### LINKED ISSUE #126: If existent, use dirid.c9r to restore filename in orphan dir fix
With #118 it is possible to get the dirid even when the ciphertext directory is orphaned.

This should be used in the orphan dir fix to additionally restore the filenames.

infeo:
Closes #126.

This PR allows to restore also filenames of an orphaned directory, if the dirId file is present.

It is implemented in a failsafe way: If the dirId file does not exists, cannot be read or does not match, the orphan cipher dir is recovered without it.

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
@@ -161,7 +161,7 @@ public class OrphanDir implements DiagnosticResult {
 	//visible for testing
 	Optional<String> retrieveDirId(Path orphanedDir, Cryptor cryptor) {
 		var dirIdFile = orphanedDir.resolve(Constants.DIR_ID_FILE);
-		var dirIdBuffer = ByteBuffer.allocate(36); //a dir id contains at most 36 ascii chars, in this impl encoded in utf8
+		var dirIdBuffer = ByteBuffer.allocate(36); //a dir id contains at most 36 ascii chars
 
 		try (var channel = Files.newByteChannel(dirIdFile, StandardOpenOption.READ); //
 			 var decryptingChannel = createDecryptingReadableByteChannel(channel, cryptor)) {
@@ -172,7 +172,7 @@ public class OrphanDir implements DiagnosticResult {
 			return Optional.empty();
 		}
 
-		var allegedDirId = StandardCharsets.UTF_8.decode(dirIdBuffer).toString();
+		var allegedDirId = StandardCharsets.US_ASCII.decode(dirIdBuffer).toString();
 
 		var dirIdHash = orphanedDir.getParent().getFileName().toString() + orphanedDir.getFileName().toString();
 		if (dirIdHash.equals(cryptor.fileNameCryptor().hashDirectoryId(allegedDirId))) {
--- a/src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
@@ -2,6 +2,7 @@ package org.cryptomator.cryptofs.health.dirid;
 
 import com.google.common.io.BaseEncoding;
 import org.cryptomator.cryptofs.CryptoPathMapper;
+import org.cryptomator.cryptofs.DirectoryIdBackup;
 import org.cryptomator.cryptofs.VaultConfig;
 import org.cryptomator.cryptofs.common.CiphertextFileType;
 import org.cryptomator.cryptofs.common.Constants;
@@ -79,7 +80,7 @@ public class OrphanDir implements DiagnosticResult {
 			return; //recovery dir was orphaned, already recovered by prepare method
 		}
 
-		var stepParentDir = prepareStepParent(dataDir, recoveryDir, cryptor.fileNameCryptor(), orphanDirIdHash);
+		var stepParentDir = prepareStepParent(dataDir, recoveryDir, cryptor, orphanDirIdHash);
 		AtomicInteger fileCounter = new AtomicInteger(1);
 		AtomicInteger dirCounter = new AtomicInteger(1);
 		AtomicInteger symlinkCounter = new AtomicInteger(1);
@@ -140,9 +141,9 @@ public class OrphanDir implements DiagnosticResult {
 	}
 
 	// visible for testing
-	CryptoPathMapper.CiphertextDirectory prepareStepParent(Path dataDir, Path cipherRecoveryDir, FileNameCryptor cryptor, String clearStepParentDirName) throws IOException {
+	CryptoPathMapper.CiphertextDirectory prepareStepParent(Path dataDir, Path cipherRecoveryDir, Cryptor cryptor, String clearStepParentDirName) throws IOException {
 		//create "step-parent" directory to move orphaned files to
-		String cipherStepParentDirName = encrypt(cryptor, clearStepParentDirName, Constants.RECOVERY_DIR_ID);
+		String cipherStepParentDirName = encrypt(cryptor.fileNameCryptor(), clearStepParentDirName, Constants.RECOVERY_DIR_ID);
 		Path cipherStepParentDirFile = cipherRecoveryDir.resolve(cipherStepParentDirName + "/" + Constants.DIR_FILE_NAME);
 		final String stepParentUUID;
 		if (Files.exists(cipherStepParentDirFile, LinkOption.NOFOLLOW_LINKS)) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -134,10 +158,53 @@ Path prepareRecoveryDir(Path pathToVault, FileNameCryptor cryptor) throws IOExce
...
+		if (dirIdHash.equals(cryptor.fileNameCryptor().hashDirectoryId(allegedDirId))) {
+			return Optional.of(allegedDirId);
+		} else {
+			LOG.info("Hash of read directory id {} does not match actual cipher dir hash {}.", allegedDirId, dirIdHash);
+			return Optional.empty();
+		}
+	}
+
+	//exists and visible for testability
+	DecryptingReadableByteChannel wrapDecryptionAround(ByteChannel channel, Cryptor cryptor) {
```
overheadhunter, line 187: I don't want to suppress any creativity, but standard names help understand method without looking at their implementation

```suggestion
	DecryptingReadableByteChannel createDecryptingReadableByteChannel(ByteChannel channel, Cryptor cryptor) {
```

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -134,10 +158,53 @@ Path prepareRecoveryDir(Path pathToVault, FileNameCryptor cryptor) throws IOExce
...
+		final String filename = filenameWithExtension.substring(0, filenameWithExtension.length() - Constants.CRYPTOMATOR_FILE_SUFFIX.length());
+		return cryptor.decryptFilename(BaseEncoding.base64Url(), filename, dirId.getBytes(StandardCharsets.UTF_8));
+	}
+
 	// visible for testing
-	void adoptOrphanedResource(Path oldCipherPath, String newClearname, CryptoPathMapper.CiphertextDirectory stepParentDir, FileNameCryptor cryptor, String longNameSuffix, MessageDigest sha1) throws IOException {
-		if (oldCipherPath.toString().endsWith(Constants.DEFLATED_FILE_SUFFIX)) {
-			var newCipherName = convertClearToCiphertext(cryptor, newClearname + longNameSuffix, stepParentDir.dirId);
+	void adoptOrphanedResource(Path oldCipherPath, String newClearname, boolean isShortend, CryptoPathMapper.CiphertextDirectory stepParentDir, FileNameCryptor cryptor, MessageDigest sha1) throws IOException {
+		var newCipherName = convertClearToCiphertext(cryptor, newClearname, stepParentDir.dirId);
```
overheadhunter, line 206: also known as "encrypt". 😄 please inline this method, as it is a single-line implementation

overheadhunter (changes requested):

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
infeo: I'll rename the method. But it is used in at least 3 different spots, hence i would keep it to reduce code duplication.

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -134,10 +158,53 @@ Path prepareRecoveryDir(Path pathToVault, FileNameCryptor cryptor) throws IOExce
...
+		try (var channel = Files.newByteChannel(dirIdFile, StandardOpenOption.READ); //
+			 var decryptingChannel = wrapDecryptionAround(channel, cryptor)) {
+			decryptingChannel.read(dirIdBuffer);
+			dirIdBuffer.flip();
+		} catch (IOException e) {
+			LOG.info("Unable to read dirIdFile of {}.", orphanedDir, e);
+			return Optional.empty();
+		}
+
+		var allegedDirId = StandardCharsets.UTF_8.decode(dirIdBuffer).toString();
```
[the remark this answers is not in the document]
infeo: It definitively increases readability, but what i like about the original approach is its resiliency: By definition the directory id is no longer than 36 ASCII chars (bytes when utf-8 encoded). Even if the dirId file is bloated with a lot of text, we don't care. and at most read 36 bytes and continue.

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
overheadhunter: I still don't agree that adding a suffix calls a method, but I won't make this a blocker for the PR

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -134,10 +158,53 @@ Path prepareRecoveryDir(Path pathToVault, FileNameCryptor cryptor) throws IOExce
...
+		}
+	}
+
+	//exists and visible for testability
+	DecryptingReadableByteChannel wrapDecryptionAround(ByteChannel channel, Cryptor cryptor) {
+		return new DecryptingReadableByteChannel(channel, cryptor, true);
+	}
+
+	//visible for testing
+	String restoreFileName(Path orphanedResource, boolean isShortened, String dirId, FileNameCryptor cryptor) throws IOException, AuthenticationFailedException {
```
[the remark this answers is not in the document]
infeo: Fixed in [62dd511](https://github.com/cryptomator/cryptofs/pull/127/commits/62dd511a97ee2f73a3234367a3cb9a1cef7d106b)

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
infeo: Fixed in [62dd511](https://github.com/cryptomator/cryptofs/pull/127/commits/62dd511a97ee2f73a3234367a3cb9a1cef7d106b)

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -134,10 +158,53 @@ Path prepareRecoveryDir(Path pathToVault, FileNameCryptor cryptor) throws IOExce
...
+
+		final String filename = filenameWithExtension.substring(0, filenameWithExtension.length() - Constants.CRYPTOMATOR_FILE_SUFFIX.length());
+		return cryptor.decryptFilename(BaseEncoding.base64Url(), filename, dirId.getBytes(StandardCharsets.UTF_8));
+	}
+
 	// visible for testing
-	void adoptOrphanedResource(Path oldCipherPath, String newClearname, CryptoPathMapper.CiphertextDirectory stepParentDir, FileNameCryptor cryptor, String longNameSuffix, MessageDigest sha1) throws IOException {
-		if (oldCipherPath.toString().endsWith(Constants.DEFLATED_FILE_SUFFIX)) {
-			var newCipherName = convertClearToCiphertext(cryptor, newClearname + longNameSuffix, stepParentDir.dirId);
+	void adoptOrphanedResource(Path oldCipherPath, String newClearname, boolean isShortend, CryptoPathMapper.CiphertextDirectory stepParentDir, FileNameCryptor cryptor, MessageDigest sha1) throws IOException {
```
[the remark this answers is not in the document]
infeo: Fixed in [62dd511](https://github.com/cryptomator/cryptofs/pull/127/commits/62dd511a97ee2f73a3234367a3cb9a1cef7d106b)

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
infeo: Changed in [62dd511](https://github.com/cryptomator/cryptofs/pull/127/commits/62dd511a97ee2f73a3234367a3cb9a1cef7d106b)

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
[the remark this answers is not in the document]
infeo: Discussion took place over a different channel, the conclusion is:
* The directory id _is_ defined (see https://docs.cryptomator.org/en/latest/security/architecture/#directory-ids)
* `ReadableByteChannel::read` is problematic, since it might not read all bytes
* cryptolib offers a [utility method](https://github.com/cryptomator/cryptolib/blob/ccfe34dc0052f906ad9c840c73f715e2ae6d4837/src/main/java/org/cryptomator/cryptolib/common/ByteBuffers.java#L46-L55) to read from a channel until either channel ends or buffers capacity is reached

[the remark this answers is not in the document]
infeo: Changed in [1d9c9b4](https://github.com/cryptomator/cryptofs/pull/127/commits/1d9c9b436377d42b455f627a709b2d193b17cb72)

I used the utility method from cryptolib.

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
```diff
@@ -134,10 +158,53 @@ Path prepareRecoveryDir(Path pathToVault, FileNameCryptor cryptor) throws IOExce
...
+		try (var channel = Files.newByteChannel(dirIdFile, StandardOpenOption.READ); //
+			 var decryptingChannel = createDecryptingReadableByteChannel(channel, cryptor)) {
+			ByteBuffers.fill(decryptingChannel, dirIdBuffer);
+			dirIdBuffer.flip();
+		} catch (IOException e) {
+			LOG.info("Unable to read dirIdFile of {}.", orphanedDir, e);
+			return Optional.empty();
+		}
+
+		var allegedDirId = StandardCharsets.UTF_8.decode(dirIdBuffer).toString();
```
overheadhunter, line 175: when you consider anything longer than 36 bytes incorrect, you shouldn't allow UTF-8 but rather use ASCII, since 36 UTF-8 chars may be 144 bytes (or even more).

overheadhunter (changes requested):

src/main/java/org/cryptomator/cryptofs/health/dirid/OrphanDir.java
infeo: Fixed in [85745fb](https://github.com/cryptomator/cryptofs/pull/127/commits/85745fb1b6c7aef559e9744924ac19879f50db5e)

overheadhunter (approved):

infeo:
I just noticed, this PR is not done yet: Since the stepparent dir is created manually, we do not add a dirid file to it, although we should. Because once the user moves the recovered directory, he won't be able to recover the files names again.

overheadhunter (approved):

PR #130, opened 2022-05-05
Include stats for metadata accesses

BeyerMart:
With this PR, two new counters are introduced, which increment, when a FileChannel is opened and when Attributes are read.

This allows to show the total amount of accesses and also provides an option to reset the second counter, so the 'Accesses per seconds' can be displayed in Cryptomator.
This feature in CryptoFs is needed for: https://github.com/cryptomator/cryptomator/issues/1511


tobihagemann:
Just wanted to chime in and confirm that newlines were reformatted. :smile: `0x0a20` (old) vs. `0x0d0a20` (new). Either set your IDE to use UNIX line endings (e.g., in IntelliJ IDEA it's in the Preferences under Editor / Code Style and then "Line separator") or you can also set it globally in your git config via `git config --global core.autocrlf true`.

overheadhunter (approved):

PR #133, opened 2022-05-12
Handle root path as filesystem API parameter gracefully

### LINKED ISSUE #132: Invoking file system methods with root path throws NullPointerException
The JDK filesystem API takes an arbitrary Path as an argument. Hence, also for the filesystem root the following is a valid method call:
```java
Path root = fs.getPath("/");
Files.createDirectory(root);
```

Although the above will likely throw an exception, it should be an `IOException`. Currently, a `NullPointerException` is thrown, caused by https://github.com/cryptomator/cryptofs/blob/59a8ec29d9a834f10a36a53202720b72020b5475/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemImpl.java#L630-L635.

We need some null-check for `Path::getFileName`. Additionally, we should check which other methods choke on the root path.

infeo:
Closes #132.

Destructive methods (`createDirectory`, `delete`, `copy` and `move`) are adjusted to throw an Filesystemexception when the root path is a parameter.

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemImpl.java
+++ b/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemImpl.java
@@ -430,6 +430,11 @@ class CryptoFileSystemImpl extends CryptoFileSystem {
 		if (cleartextSource.equals(cleartextTarget)) {
 			return;
 		}
+
+		if (rootPath.equals(cleartextTarget) && ArrayUtils.contains(options, StandardCopyOption.REPLACE_EXISTING)) {
+			throw new FileSystemException("The filesystem root cannot be replaced.");
+		}
+
 		CiphertextFileType ciphertextFileType = cryptoPathMapper.getCiphertextFileType(cleartextSource);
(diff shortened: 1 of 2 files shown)


src/main/java/org/cryptomator/cryptofs/CryptoFileSystemImpl.java
```diff
@@ -515,6 +529,15 @@ private void copyAttributes(Path src, Path dst) throws IOException {
 	void move(CryptoPath cleartextSource, CryptoPath cleartextTarget, CopyOption... options) throws IOException {
 		readonlyFlag.assertWritable();
 		assertCleartextNameLengthAllowed(cleartextTarget);
+
+		if (rootPath.equals(cleartextSource)) {
+			throw new FileSystemException("Filesystem root cannot be deleted.");
```
overheadhunter, line 534:
```suggestion
			throw new FileSystemException("Filesystem root cannot be moved.");

```

overheadhunter (changes requested):
Looks reasonable, just found a little copy-paste error.

PR #135, opened 2022-05-25
Add "missing dirid backup file" result to directory check

### LINKED ISSUE #128: Add DirectoryCheck result indicating the dirId file is missing
With #127 merged, it is benefical for a cipher directory to have a dirid file increasing the overall resiliency of the vault structure against data loss.

But in the preceding #118, we decided to not automatically migrate already existing directories in a vault. To allow the user to still benefit from the change, we can add a directory check result `MissingDirIdFile` with INFO error level and the fix to add such file.

infeo:
Closes #128.

The result has severity `WARN`. 
The fix is to create the backup file.
The result is created, when a healthy Dir without the backup file is found.


overheadhunter (approved):

infeo:
For the record: The build fails not due to test errors, but because the maven-dependency-plugin has some errors 'gain.

PR #138, opened 2022-07-15
More dirIdCheck results

infeo:
This PR adds two new results to the DirIdCheck:
1. LooseDirIdFile: Every file named `dir.c9r` which parent does not end with `.c9r`. The fix is to remove the file.
2. MissingDirIdFile: Every c9r directory, which does not contain a dir.c9r file The fix is to remove the directory to not hog a name.

The second is especially interesing in combination of the oprhan dir result. The orphaned directory can be recovered and the old c9r dir formerly containing the link file can be removed, such that the user can add the directory at its original location again.

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/health/dirid/LooseDirFile.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/dirid/LooseDirFile.java
@@ -12,17 +12,22 @@ import java.util.Map;
 
 import static org.cryptomator.cryptofs.health.api.CommonDetailKeys.DIR_ID_FILE;
 
-public class LooseDirIdFile implements DiagnosticResult {
+public class LooseDirFile implements DiagnosticResult {
 
 	final Path dirFile;
 
-	LooseDirIdFile(Path dirFile) {
+	LooseDirFile(Path dirFile) {
 		this.dirFile = dirFile;
 	}
 
 	@Override
 	public Severity getSeverity() {
 		return Severity.INFO;
+	}
+
+	@Override
+	public String toString() {
+		return String.format("A dir.c9r without proper parent found: (%s). .", dirFile);
 	}
 
 	@Override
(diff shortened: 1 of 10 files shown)


src/main/java/org/cryptomator/cryptofs/health/dirid/LooseDirFile.java
```diff
@@ -0,0 +1,42 @@
...
+import org.cryptomator.cryptolib.api.Masterkey;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Map;
+
+import static org.cryptomator.cryptofs.health.api.CommonDetailKeys.DIR_ID_FILE;
+
+public class LooseDirFile implements DiagnosticResult {
```
overheadhunter, line 15: Please keep the convention of calling this a `DirIdFile`, since it always has been this way and causes confusion to rename it.

src/main/java/org/cryptomator/cryptofs/health/dirid/MissingDirFile.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Map;
+
+/**
+ *  A c9r directory without a dirId file.
+ */
+public class MissingDirFile implements DiagnosticResult {
```
overheadhunter, line 17: Same. It is not a file containing a missing dir, but it is the dir id file that is missing.

overheadhunter (changes requested):

src/main/java/org/cryptomator/cryptofs/health/dirid/LooseDirFile.java
infeo: Changed in [82775e7](https://github.com/cryptomator/cryptofs/pull/138/commits/82775e7e4fce91350a0b461bf4512126a2b88b42)

src/main/java/org/cryptomator/cryptofs/health/dirid/MissingDirFile.java
infeo: Changed in [82775e7](https://github.com/cryptomator/cryptofs/pull/138/commits/82775e7e4fce91350a0b461bf4512126a2b88b42)

infeo:
I also renamed some other classes to align those with the naming convention

overheadhunter (approved):

overheadhunter (approved):

PR #139, opened 2022-07-20
Implement fix for missing content dir

infeo:
The fix for a missing content dir is to create the missing directory and place a dirId backup file inside.

Creating the directory again can cause sync conflict, but not fixing the result reserves a resource name in the parent directory unnoticable for the library consumer. Since the health check is for fixing problems with the present vault structure, the latter rules out the former from my point of view.


src/main/java/org/cryptomator/cryptofs/health/dirid/MissingContentDir.java
```diff
@@ -0,0 +1,60 @@
...
+		return Map.of(DIR_ID, dirId, //
+				DIR_ID_FILE, file.toString());
+	}
+
+	@Override
+	public void fix(Path pathToVault, VaultConfig config, Masterkey masterkey, Cryptor cryptor) throws IOException {
+		var dirIdHash = cryptor.fileNameCryptor().hashDirectoryId(dirId);
+		Path dirPath = pathToVault.resolve(Constants.DATA_DIR_NAME).resolve(dirIdHash.substring(0, 2)).resolve(dirIdHash.substring(2, 30));
+		Files.createDirectories(dirPath);
+		createDirIdFile(cryptor, new CryptoPathMapper.CiphertextDirectory(dirId, dirPath));
```
overheadhunter, line 53: It took me some time to understand that the `DirId` file is not the `dir` file (which happens to also contain the dir ID). While I'm fine with the file name (since it is basically the same - but with encryption in place), we might want to establish a different name in code (e.g. `createDirIdBackupFile()`)...

overheadhunter (approved):

src/main/java/org/cryptomator/cryptofs/health/dirid/MissingContentDir.java
infeo: Thanks for the feedback, i renamed the method in https://github.com/cryptomator/cryptofs/pull/139/commits/14f5c96487a34237ed68e4d3f87c7baaca1aa333

PR #140, opened 2022-09-09
fix `CryptoBasicFileAttributes` #GH-141

revintec:
the old implementation returns wrong file size for symbolic links, e.g. file `symlink` after `ln -s . symlink` should be 1


overheadhunter:
Hi, can you please:
1. Create a bug report claiming that symlink size is incorrect and provide some official source documenting what the correct size of a symlink should be like?
2. When "fixing" the symlink size, please don't break the directory size 😉

overheadhunter (changes requested):
Thanks for creating the corresponding issue. The change looks reasonable, however I would like to ask you to stay with the switch statement, as with the ternary operator we lose compile-time exhaustiveness checks.

Furthermore simply moving `SYMLINK` from one case to the other introduces a significantly smaller diff.

overheadhunter (approved):

overheadhunter:
Thank you very much!

PR #144, opened 2022-10-06
support atomic open + set attributes

### LINKED ISSUE #143: file attributes ignored when creating new file channel
Note that the `attrs` in this method are not passed down to `OpenCryptoFile#newFileChannel`:

https://github.com/cryptomator/cryptofs/blob/c5432f87b557408cca2a286f664ea6effbf931a6/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemImpl.java#L363-L395

https://github.com/cryptomator/cryptofs/blob/c5432f87b557408cca2a286f664ea6effbf931a6/src/main/java/org/cryptomator/cryptofs/fh/OpenCryptoFile.java#L68-L111

Which leads to atomic "create and set attributes" to fail. Here is an integration test showing the problem:

```java
import java.nio.channels.*;
import java.nio.file.*;
import java.nio.file.attribute.*;

var path = Path.of("/Volumes/foo/test.txt");
var oo = Set.of(StandardOpenOption.WRITE, StandardOpenOption.CREATE_NEW);
var attr = PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rwxr-x---"));

var fc = FileChannel.open(path, Set.of(StandardOpenOption.WRITE, StandardOpenOption.CREATE_NEW), attr);
fc.write(java.nio.charset.StandardCharsets.UTF_8.encode("hello world"));
fc.close();
```

overheadhunter:
This fixes #143 by passing through the `attrs` part of [`FileSystemProvider.html#newFileChannel(path, openOptions, attrs)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/spi/FileSystemProvider.html#newFileChannel(java.nio.file.Path,java.util.Set,java.nio.file.attribute.FileAttribute...)) to the ciphertext file channel.

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemImpl.java
+++ b/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemImpl.java
@@ -378,7 +378,7 @@ class CryptoFileSystemImpl extends CryptoFileSystem {
 			Files.createDirectories(ciphertextPath.getRawPath()); // suppresses FileAlreadyExists
 		}
 
-		FileChannel ch = openCryptoFiles.getOrCreate(ciphertextFilePath).newFileChannel(options); // might throw FileAlreadyExists
+		FileChannel ch = openCryptoFiles.getOrCreate(ciphertextFilePath).newFileChannel(options, attrs); // might throw FileAlreadyExists
 		try {
 			if (options.writable()) {
 				ciphertextPath.persistLongFileName();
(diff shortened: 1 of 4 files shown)


overheadhunter:
Fix confirmed via fuse-nio-adapter mirror with a cryptofs snapshot build and the jshell command described in #143

tobihagemann (approved):

PR #149, opened 2022-11-15
Health API: Decouple result fix from severity

infeo:
Currently, the (undocumented) convention for an implementaiton of `DiagnosticResult` is the following:
result severity  != `WARN` → do not override `fix()` method. (aka throws `UnsupportedOperationException`).

In the context of https://github.com/cryptomator/cryptofs/issues/128, this convention showed a lack of flexibility. The dirId backup file is an optional feature, but the result stating it's missing has a severity of `WARN` to be fixable.
Additionally, a lib consumer is unable to reach an intact vault structure once a critical error appears. This leads to reserved file paths, undeletable resources and other permament errors.

This PR solves those problems by discarding the aforementioned convention. The ideas are
* give the consumer the information she needs to make decisions
* provide tools to fix the vault structure and make the filesystem error free again
* only consider the current local state of the vault

The ideas are realized by decoupling the severity and the fix() method of a `DiagnosticResult` and change their meaning:
* The severity serves as an _indicator if the vault structure is compromised or data loss happened_. It does not make statements about if the result can be fixed/resolved.
* The fix method exists to _heal the vault structure_. More technically speaking,  it resolves the result by altering files and directories, such that the result is not triggered anymore by a subsequent run of the same health check. 

Implementationwise, the javadoc of `Severity` is adjusted. The feature to determine if a result fix exists is moved from convention to implementation: `DiagnosticResul::getFix` is a new method which returns `Optional<Fix>`, where the latter class is just a functional interface to apply the fix. The default implementation of `::getFix` returns an empty optional. To keep compatibility, method `fix(...)` still exists, but is deprecated. For existing results with fixes, a delegate method is implemented.

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/health/api/DiagnosticResult.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/api/DiagnosticResult.java
@@ -63,7 +63,9 @@ public interface DiagnosticResult {
 	 */
 	@Deprecated
 	default void fix(Path pathToVault, VaultConfig config, Masterkey masterkey, Cryptor cryptor) throws IOException {
-		throw new UnsupportedOperationException("Fix for result" + this.getClass() + " not implemented");
+		getFix(pathToVault, config, masterkey, cryptor) //
+				.orElseThrow(() -> new UnsupportedOperationException("Fix for result" + this.getClass() + " not implemented")) //
+				.apply();
 	}
 
 	default Optional<Fix> getFix(Path pathToVault, VaultConfig config, Masterkey masterkey, Cryptor cryptor) {
@@ -81,6 +83,7 @@ public interface DiagnosticResult {
 
 	@FunctionalInterface
 	interface Fix {
+
 		void apply() throws IOException;
 	}
 }
--- a/src/main/java/org/cryptomator/cryptofs/health/api/DiagnosticResult.java
+++ b/src/main/java/org/cryptomator/cryptofs/health/api/DiagnosticResult.java
@@ -7,6 +7,7 @@ import org.cryptomator.cryptolib.api.Masterkey;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Map;
+import java.util.Optional;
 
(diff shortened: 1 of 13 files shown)


infeo:
Remark: To keep the diff small, this branch merges into https://github.com/cryptomator/cryptofs/tree/feature/missing-c9rcontent-is-type. There is an open PR for that one: https://github.com/cryptomator/cryptofs/pull/148

Once merge, the merge target should be changed to `develop`.

src/main/java/org/cryptomator/cryptofs/health/api/DiagnosticResult.java
```diff
@@ -40,10 +48,28 @@ enum Severity {
...
+	 * @param masterkey the masterkey of the vault
+	 * @param cryptor
+	 * @throws IOException
+	 * @throws UnsupportedOperationException if no fix is implemented for this result
+	 * @deprecated Use {@link #getFix(Path, VaultConfig, Masterkey, Cryptor)} instead
+	 */
+	@Deprecated
 	default void fix(Path pathToVault, VaultConfig config, Masterkey masterkey, Cryptor cryptor) throws IOException {
 		throw new UnsupportedOperationException("Fix for result" + this.getClass() + " not implemented");
 	}
```
overheadhunter, line 67: Maybe only throw UOE off the Optional is empty? Then implementations can drop this method, but API consumers don't notice a change.

overheadhunter (changes requested):

src/main/java/org/cryptomator/cryptofs/health/api/DiagnosticResult.java
infeo: Implemented in https://github.com/cryptomator/cryptofs/pull/149/commits/3ce839049e80814dd617095b42a4771695a0e243.

I left all fix() method in the result impls for testing, but changed their signature (visibility and parameters).

overheadhunter (approved):

PR #150, opened 2022-11-20
Health check: Get affected fs nodes of result

infeo:
This PR adds the method `getCausingCiphertextNodes()` to the interface `DiagnosticResult`.

The method is intended to list ciphertext resources, which triggered the concrete result.


infeo:
I openend https://github.com/cryptomator/cryptofs/issues/151 to discuss any possible solutions for fix resolving

PR #154, opened 2022-12-30
Feature: Add function to get (data) ciphertext path from cleartext path

infeo:
This PR adds to the CryptoFilesystem interface the following function: 
`Path getPathToDataCiphertext(Path cleartextPath) throws IOException`

The function returns for a given, absolute cleartext path belonging to this cryptofilesystem its ciphertext data counterpart. It does not point to intermediante files like dir.c9r. This means:
* regular file `/file.file` → `/d/AB/CD...XYZ/foo.c9r`
* shortend file `/very...long.name` → `/d/AB/CD...XYZ/foo.c9s/contents.c9r`
* directory `/directory` → `/d/AB/CD...XYZ/`
* symlink `/sym.link` → `/d/AB/CD...XYZ/foo.c9r/symlink.c9r`

If the fs node does not exists or another exception is thrown, it is propageted.

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/CryptoFileSystem.java
+++ b/src/main/java/org/cryptomator/cryptofs/CryptoFileSystem.java
@@ -1,5 +1,6 @@
 package org.cryptomator.cryptofs;
 
+import java.io.IOException;
 import java.nio.file.FileSystem;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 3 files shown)


src/main/java/org/cryptomator/cryptofs/CryptoFileSystem.java
```diff
@@ -29,6 +30,15 @@ public abstract class CryptoFileSystem extends FileSystem {
...
 
+	/**
+	 * Provides the {@link Path} to the (data) ciphertext from a given cleartext path.
+	 *
+	 * @param cleartextPath absolute path to the cleartext file or folder belonging to this {@link CryptoFileSystem}. Internally the path must be an instance of {@link CryptoPath}
+	 * @return the {@link Path} to ciphertext file or folder containing teh actual encrypted data
+	 * @throws IOException
+	 */
+	public abstract Path getPathToDataCiphertext(Path cleartextPath) throws IOException;
+
```
overheadhunter, line 41: 1. let's call this method `getCiphertextPath` (if it wasn't for directories, I would even call it `getC9rPath`)
2. document known specific exceptions, such as `ProviderMismatchException`
3. add description to `@throws IOException`

overheadhunter (commented):
Looks good, I just want to rename the method and enhance docs

overheadhunter (approved):

src/main/java/org/cryptomator/cryptofs/CryptoFileSystem.java
infeo: Resolved in https://github.com/cryptomator/cryptofs/pull/154/commits/8167d5d6dc60b36d1f7ee6baefb489c7cbdccf8e

PR #156, opened 2023-01-05
Support for encrypted UserDefinedAttributeView

### LINKED ISSUE #47: Support for UserDefinedFileAttributeView
If the underlying file system [supports it](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/FileStore.html#supportsFileAttributeView(java.lang.Class)), we should allow retrieval of a [`UserDefinedFileAttributeView`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/attribute/UserDefinedFileAttributeView.html).

Maximum xattr name length on macOS is 127 chars (255 on Linux, unknown on Windows). Subtracting the prefix, we have 118 usable chars to store a base64-encoded value or in other words: 87 bytes. If we subtract a 16 byte IV, we can store up to 71 cleartext bytes.

Therefore we need to decide, if this is sufficient or whether we want a deterministic scheme (SIV) and then hash the name, which slows down listing but essentially allows any length. We need to [collect some data](https://community.cryptomator.org/t/how-long-are-your-extended-attributes/7280) to decide this.

One further advantage of using SIV (+hash), is the ability to cache names and thus speed up xattr listings and decryption independent of the dir. This comes with the disadvantage, that xattr names repeat across different dirs, leaking some information.

Content encryption should be the same as for file contents.

overheadhunter:
Fixes #47

* Attribute names are `c9r.` + base64url-encoded siv-encrypted attribute names
* Attribute values are encrypted in the same way as file contents

There is no shortening. TBD: How to deal with too long attribute names?

This PR replaces the stale branch [feature/47-user-attr](https://github.com/cryptomator/cryptofs/tree/feature/47-user-attr).

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/attr/AbstractCryptoFileAttributeView.java
+++ b/src/main/java/org/cryptomator/cryptofs/attr/AbstractCryptoFileAttributeView.java
@@ -20,7 +20,7 @@ import java.nio.file.Path;
 import java.nio.file.attribute.FileAttributeView;
 
 abstract sealed class AbstractCryptoFileAttributeView implements FileAttributeView
-		permits CryptoBasicFileAttributeView, CryptoFileOwnerAttributeView, CryptoUserDefinedAttributeView {
+		permits CryptoBasicFileAttributeView, CryptoFileOwnerAttributeView, CryptoUserDefinedFileAttributeView {
 
 	protected final CryptoPath cleartextPath;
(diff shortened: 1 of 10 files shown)


infeo (commented):
Looks good.

Regarding handling of too long attributes: I'm totally fine by defining a limit and throw an IOException if the limit is exceeded. But I honstely don't know, what an appropiate limit would be.

PR #159, opened 2023-03-15
Feature: Use dagger factories instead of builders

infeo:
This PR replaces all dagger `@Subcomponent.Builder`s with dagger [Factories](https://dagger.dev/api/latest/dagger/Component.Factory.html).

Factories guarentee to always return a new instance and do not have a public state, hence they are threadsafe.

### DIFF
--- a/src/test/java/org/cryptomator/cryptofs/fh/OpenCryptoFilesTest.java
+++ b/src/test/java/org/cryptomator/cryptofs/fh/OpenCryptoFilesTest.java
@@ -16,12 +16,12 @@ import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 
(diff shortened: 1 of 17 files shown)


src/test/java/org/cryptomator/cryptofs/fh/OpenCryptoFilesTest.java
```diff
@@ -16,12 +16,12 @@
 import java.nio.file.Path;
 import java.nio.file.Paths;
 
+import static org.mockito.ArgumentMatchers.any;
```
overheadhunter, line 19: Wrong import

overheadhunter (approved):
lgtm

src/test/java/org/cryptomator/cryptofs/fh/OpenCryptoFilesTest.java
infeo: Fixed in https://github.com/cryptomator/cryptofs/pull/159/commits/62d7e9edf0c25d784983ce84a798c01369b84674

Issue #158, opened 2023-03-14
Chunk ByteBuffer to soon recycled


infeo:
In PR https://github.com/cryptomator/cryptofs/pull/125, the class `BufferPool` was introduced to reduce buffer allocations and reuse already allocated buffers for de- and encryption. The affected buffers are part of "Chunk" records, which represent a contigous memory area in a file.

A buffer is recylced, when:
* a chunk is saved to the ciphertextfile
* a chunk is loaded from the ciphertextfile and
* a chunk is evicted from the `ChunkCache`:
https://github.com/cryptomator/cryptofs/blob/ec170e76a75dc3188e32576a5b3326a2404f74fd/src/main/java/org/cryptomator/cryptofs/fh/ChunkCache.java#L47-L54

The third point (recycle on eviction) causes a sever bug. Imagine the following scenario:
* Small size of chunk cache (e.g., 5)
* Great number of concurrent reading threads (e.g., 100)
* File spanning over several chunks
* Random reading pattern leading to a low chunk cache hit rate

Low hit rate -> High number of cache evictions -> High recylce rate. But only because a chunk is evicted does not imply, the thread reading on the chunk has finished its operation. If the recycled buffer is selected to be filled again, the (still not finished) thread is reading wrong content.

infeo:
To recycle the data buffer of Chunks, a synchronization mechanism is needed which ensures:
1. Once the chunk is evicted from cache, no new accesses to data can be made
2. the chunk keeps track of accesses
3. Only if the last access is _"closed"_, the buffer is recycled

This would add another layer of complexity, hence, for now we decided to keep it simple and just synchronize threads on methods in `CleartextFileChannel`.

PR #161, opened 2023-03-22
refactor chunk cache

overheadhunter:
This is a new attempt of fixing #158 while retaining the ability to concurrently read from the same open file.

The ChunkCache will now make sure not to recycle any chunk as long as it is still being used ("active chunk" - using reference counting and AutoCloseable). When no longer active, chunks are moved to a "stale" cache, invalidating LRU chunks eventually.

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/fh/ChunkCache.java
+++ b/src/main/java/org/cryptomator/cryptofs/fh/ChunkCache.java
@@ -60,6 +60,12 @@ public class ChunkCache {
 	 */
 	public Chunk putChunk(long chunkIndex, ByteBuffer chunkData) throws IllegalArgumentException {
 		return activeChunks.compute(chunkIndex, (index, chunk) -> {
+			// stale chunk for this index is obsolete:
+			var staleChunk = staleChunks.remove(index);
+			if (staleChunk != null) {
+				bufferPool.recycle(staleChunk.data());
+			}
+			// either create completely new chunk or replace all data of existing active chunk:
 			if (chunk == null) {
 				chunk = new Chunk(chunkData, true, () -> releaseChunk(chunkIndex));
 			} else {
--- a/src/main/java/org/cryptomator/cryptofs/fh/ChunkCache.java
+++ b/src/main/java/org/cryptomator/cryptofs/fh/ChunkCache.java
@@ -2,6 +2,7 @@ package org.cryptomator.cryptofs.fh;
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.RemovalCause;
 import com.google.common.cache.RemovalNotification;
 import org.cryptomator.cryptofs.CryptoFileSystemStats;
 import org.cryptomator.cryptolib.api.AuthenticationFailedException;
@@ -9,7 +10,8 @@ import org.cryptomator.cryptolib.api.AuthenticationFailedException;
 import javax.inject.Inject;
 import java.io.IOException;
 import java.io.UncheckedIOException;
-import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 @OpenFileScoped
 public class ChunkCache {
@@ -20,7 +22,8 @@ public class ChunkCache {
 	private final ChunkSaver chunkSaver;
 	private final CryptoFileSystemStats stats;
 	private final BufferPool bufferPool;
-	private final Cache<Long, Chunk> chunks;
+	private final Cache<Long, Chunk> staleChunks;
+	private final ConcurrentMap<Long, Chunk> activeChunks;
 
 	@Inject
 	public ChunkCache(ChunkLoader chunkLoader, ChunkSaver chunkSaver, CryptoFileSystemStats stats, BufferPool bufferPool) {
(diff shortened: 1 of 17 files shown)


src/main/java/org/cryptomator/cryptofs/fh/ChunkCache.java
```diff
@@ -20,54 +25,166 @@ public class ChunkCache {
...
-		stats.addChunkCacheMiss();
+	/**
+	 * Overwrites data at the given index and increments the access counter of the returned chunk
+	 *
+	 * @param chunkIndex Which chunk to overwrite
+	 * @param chunkData The cleartext data
+	 * @return The chunk
+	 * @throws IllegalArgumentException If {@code chunkData}'s remaining bytes is not equal to the number of bytes fitting into a chunk
+	 */
+	public Chunk putChunk(long chunkIndex, ByteBuffer chunkData) throws IllegalArgumentException {
```
infeo, line 61: If we call `putChunk` for an index, which is also mapped in `staleChunks`, that index is present in _both_ maps (active and stale). In my opinion, we should disallow such states and ensure the following invariant:
> At any point in time, an index is mapped either in active or stale, but not in both.

Otherwise we get cases like
1. Index 42 is only mapped to chunk c1 (42:c1) in `staleChunks`
1. Thread t_A calls `getChunk(42)`, moving (42:c1) from stale to active
1. Thread t_B calls `putChunk(42,data)`, replacing (42:c1) with (42:c2) in active
1. Thread t_A closes  c1, which leads to:
    - calling `releaseChunk(42)`
    - decrement the access counter of _c2_ (a different chunk!)
    - moving (42:c2) from active to stale
1.  Chunk c1 is "lost", its buffer won't be recylced and is only removed in the next gc run
1. Thread t_B closes c2 (because it has a reference to it), which will throw a NullPointerException, because chunk is not in active anymore

src/test/java/org/cryptomator/cryptofs/fh/ChunkCacheTest.java
```diff
@@ -20,147 +25,170 @@ public class ChunkCacheTest {
...
-		long indexNotInCache = 40L;
-		Chunk chunk = mock(Chunk.class);
-		inTest.set(firstIndex, chunk);
-		for (int i = 1; i < ChunkCache.MAX_CACHED_CLEARTEXT_CHUNKS; i++) {
-			inTest.set(firstIndex + i, mock(Chunk.class));
-		}
-		when(chunkLoader.load(indexNotInCache)).thenReturn(mock(Chunk.class));
-
-		inTest.get(indexNotInCache);
+	@DisplayName("putChunk returns a dirty chunk")
```
infeo, line 84: We should also have a test "putChunk returns chunk with accessCounter > 0

src/test/java/org/cryptomator/cryptofs/fh/ChunkCacheTest.java
```diff
@@ -20,147 +25,170 @@ public class ChunkCacheTest {
...
-		verify(stats).addChunkCacheAccess();
-		verify(chunkSaver).save(firstIndex, chunk);
-		verify(bufferPool).recycle(chunk.data());
-		verifyNoMoreInteractions(chunkSaver);
+		Assertions.assertTrue(chunk.isDirty());
 	}
 
-	@Test
-	public void testGetInvokesSaverIfMaxEntriesInCacheAreReachedAndAnEntryNotInCacheIsSet() throws IOException {
+	@RepeatedTest(100)
```
infeo, line 93: Why do we need a repeated test here?

src/test/java/org/cryptomator/cryptofs/fh/ChunkCacheTest.java
```diff
@@ -20,147 +25,170 @@ public class ChunkCacheTest {
...
+			inTest.flush();
+
+			verify(chunkSaver).save(Mockito.eq(index), Mockito.any());
+			verify(chunkSaver).save(Mockito.eq(index2), Mockito.any());
+		}
 	}
 
 	@Test
-	public void testGetRethrowsIOExceptionFromLoader() throws IOException, AuthenticationFailedException {
+	@DisplayName("flush() saves all stale chunks")
```
infeo, line 167: Just an idea: A test, which ensures that flush() does not change active and stale.

infeo (changes requested):
I love the new architecture, where a chunk is an autoclosable object.

But there are still some open question and not-tested branches.

src/test/java/org/cryptomator/cryptofs/fh/ChunkCacheTest.java
overheadhunter: With the old implementation I noticed that it failed on CI but succeeded on my local machine. I then noticed that the implementation was timing-sensitive: If you remove the `cdl.await()`, you'll see that `chunkSaver.save(...)` is not called immediately but in an async maintenance thread (as mentioned in the Caffeine docs). Therefore the verifications in this tests failed _sometimes_.

src/main/java/org/cryptomator/cryptofs/fh/ChunkCache.java
overheadhunter: I don't think this is an issue. Step 2 and 3 are each atomic. So regardless of which thread runs first, either `c1` xor `c2` will be in `activeChunks`.

src/test/java/org/cryptomator/cryptofs/fh/ChunkCacheTest.java
overheadhunter: I added assertions for the access count in 32f0ef4:

https://github.com/cryptomator/cryptofs/blob/32f0ef4b93012bd0bae0ec90a5f0a2a639f60749/src/test/java/org/cryptomator/cryptofs/fh/ChunkCacheTest.java#L244-L269

src/main/java/org/cryptomator/cryptofs/fh/ChunkCache.java
```diff
@@ -20,54 +25,166 @@ public class ChunkCache {
...
+	 * @see #invalidateAll()
+	 */
+	public void flush() throws IOException {
+		var lock = flushLock.writeLock();
+		lock.lock();
+		BiFunction<Long, Chunk, Chunk> saveUnchecked = (index, chunk) -> {
+			try {
+				chunkSaver.save(index, chunk);
+			} catch (IOException e) {
+				throw new UncheckedIOException(e);
```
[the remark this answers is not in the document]
overheadhunter: I tested it and it turns out the mapping stays unchanged in case of an exception. Since this behaviour is not documented, I decided to add a test in 32f0ef4 to make sure we notice it, should it ever change:

https://github.com/cryptomator/cryptofs/blob/32f0ef4b93012bd0bae0ec90a5f0a2a639f60749/src/test/java/org/cryptomator/cryptofs/fh/ChunkCacheTest.java#L211-L221

src/main/java/org/cryptomator/cryptofs/fh/ChunkCache.java
ben-manes: fwiw, you could use one mapping by [pinning](https://github.com/ben-manes/caffeine/wiki/Faq#pinning-entries) so that active chunks have a weight of zero. I don’t know if it simplifies your logic, though.

overheadhunter: That would indeed simplify things! I guess it should be a separate PR though 😅

infeo: > either c1 xor c2 will be in activeChunks.

True, my mistake. Nevertheless, we discussed this in a different channel and decided to ensure this invariant.

> could use one mapping by [pinning](https://github.com/ben-manes/caffeine/wiki/Faq#pinning-entries) so that active chunks have a weight of zero

Then we would only need one map actually and don't have to move around chunks, which simpliyfies our architecture. @ben-manes Thanks for pointing that out!

infeo (approved):

PR #163, opened 2023-03-23
Simplified ChunkCache

overheadhunter:
As suggested during review of #161, using weights on a Caffeine cache it is possible [to pin active chunks](https://github.com/ben-manes/caffeine/wiki/Faq#pinning-entries), avoiding premature eviction.

This simplifies our ChunkCache by purely relying on the atomic operations of said cache and avoids moving chunks between an `active` and `stale` collection.


infeo (changes requested):
Phew, had to think hard about this PR, but i think we are still missing one thing:

Let's say, currently runs thread t1, in which a chunk is evicted and saved. Before t1 executes `ciphertext.write(encryptedChunkData)` in `ChunkSaver`, it is put on hold.
Simultaneously, thread t2 opens a new file channel with ´TRUNCATE_EXISTING`. (meaning cache is invalidated and ciphertext truncated). t2 finishes the job, t1 resumes and actually writes content beyond the truncation.

Actual result: File is not truncated. (t1 wrote chunk)
Imho expected result: File is truncated (because caches should be ignored or cleared)

Meaning, when we truncate the ciphertext, we not only need to invalidate the cache, but also cancel/abort all non-active writes (e.g. by ChunkSaver).

overheadhunter:
> Since all mentioned methods obtain a lock [...]

I was mistaken: Those methods hold a lock, but eviction runs on a different thread. That is, unless [you tell it otherwise](https://www.javadoc.io/doc/com.github.ben-manes.caffeine/caffeine/latest/com.github.benmanes.caffeine/com/github/benmanes/caffeine/cache/Caffeine.html):

> The executor is delegated to when [...] performing periodic maintenance.

By setting the executor to `Runnable::run`, we can make sure to run maintenance (i.e. eviction) directly on the same thread, making the eviction behaviour more coherent and comprehensible for our purpose.

PR #164, opened 2023-03-24
migrated from Guava Cache to Coffeine

### LINKED ISSUE #162: Replace Guava Caches with Caffeine
Caffeine is a modern cache-library, which is based on Guavas caches API but with improvements and (according to themself) a near optimal cache eviction strategy.

The first step by defining it as dependency was made in #161. We should replace all remaining occurrences of the Guavas `CacheBuilder` with the cache builder provided by Caffeine.

overheadhunter:
fixes #162

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/CryptoPathMapper.java
+++ b/src/main/java/org/cryptomator/cryptofs/CryptoPathMapper.java
@@ -8,11 +8,9 @@
  *******************************************************************************/
 package org.cryptomator.cryptofs;
 
-import com.google.common.base.Throwables;
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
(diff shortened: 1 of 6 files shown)


infeo (approved):

infeo:
I gave this PR another look and noticed, that we have different handling of the UncheckedIOExceptions:
1. Rewrap it in a generic `IOException`, loosing direct type information about what failed (e.g., https://github.com/cryptomator/cryptofs/pull/164/files#diff-2261d9d07dd9f059e74785d49eb7ed731d267b4f53eb030617ac116fe22b9ad6R161-R170)
1. Unwrap and rethrow the cause, as done in https://github.com/cryptomator/cryptofs/pull/164/files#diff-bda48bb261f4b28cb28912d63eaffd3f12979360757e812b8f45784be17ac850L73-L80

@overheadhunter What is the logic behind it?

Issue #171, opened 2023-04-14
Certain operations are not atomic in respect to the lib consumer


infeo:
As the library consumer, i expect, that no (access, timing)-pattern creates an invalid vault structure (TODO: define what this means).

PR #172, opened 2023-04-14
Feature: Remove deleted file from openFiles map and set file path to null

### LINKED ISSUE #170: Creating file channel to previously deleted file might contain cached content
Currently, when [FileSystemProvider.delete(Path p)](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/spi/FileSystemProvider.html#delete(java.nio.file.Path)) is called for a file in cryptoFS, just the encrypted file is deleted:
https://github.com/cryptomator/cryptofs/blob/b3755f354e4d57d47ab655cf43e8d83c41685c25/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemImpl.java#L415-L426

There is no cache invalidation for the path or openCryptoFiles. This can lead to reading deleted content:
```java
@Test
public void testCreateNewThenWriteThenDeleteThenRead() throws IOException {
	var bufToWrite = ByteBuffer.wrap("delete me".getBytes(StandardCharsets.UTF_8));
	final int bytesRead;
	try (var ch = FileChannel.open(file, CREATE_NEW, WRITE)) {
		ch.write(bufToWrite);
		Files.delete(file);
		try (var ch2 = fileSystem.provider().newFileChannel(file, Set.of(CREATE, READ, WRITE))) {
			bytesRead = ch2.read(ByteBuffer.allocate(bufToWrite.capacity()));
		}
	}
	Assertions.assertEquals(-1, bytesRead);
}
```

infeo:
Fixes #170.

Before actually deleting files on the underlying file system, the openCryptoFile is removed from the (path, file)-mapping. Additionally, the `currentFilePath` object is null'ed, to indicate the openCryptoFile should not be used anymore. The usages of `currentFilePath` are updated to be null safe.

Note, that this only fixes a "symptom", there is still surface for bad timing. The root cause is described in #171.

### DIFF
--- a/src/test/java/org/cryptomator/cryptofs/CryptoFileChannelWriteReadIntegrationTest.java
+++ b/src/test/java/org/cryptomator/cryptofs/CryptoFileChannelWriteReadIntegrationTest.java
@@ -561,7 +561,7 @@ public class CryptoFileChannelWriteReadIntegrationTest {
 		//https://github.com/cryptomator/cryptofs/issues/170
 		@Test
 		public void testWriteThenDeleteThenRead() throws IOException {
-			var bufToWrite = ByteBuffer.wrap("delete me".getBytes(StandardCharsets.UTF_8));
+			var bufToWrite = StandardCharsets.UTF_8.encode("delete me");
 			final int bytesRead;
 			try (var ch = FileChannel.open(file, CREATE_NEW, WRITE)) {
 				ch.write(bufToWrite);
--- a/src/test/java/org/cryptomator/cryptofs/CryptoFileChannelWriteReadIntegrationTest.java
+++ b/src/test/java/org/cryptomator/cryptofs/CryptoFileChannelWriteReadIntegrationTest.java
@@ -48,6 +48,7 @@ import java.time.temporal.ChronoUnit;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Random;
+import java.util.Set;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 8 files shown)


src/test/java/org/cryptomator/cryptofs/CryptoFileChannelWriteReadIntegrationTest.java
```diff
@@ -556,6 +557,21 @@ public void testClosingChannelOfDeletedFileDoesNotThrow() {
 			});
 			Assertions.assertTrue(Files.notExists(file));
 		}
+
+		//https://github.com/cryptomator/cryptofs/issues/170
+		@Test
+		public void testWriteThenDeleteThenRead() throws IOException {
+			var bufToWrite = ByteBuffer.wrap("delete me".getBytes(StandardCharsets.UTF_8));
```
overheadhunter, line 564:
```suggestion
			var bufToWrite = StandardCharsets.UTF_8.encode("delete me");

```

overheadhunter (approved):

src/test/java/org/cryptomator/cryptofs/CryptoFileChannelWriteReadIntegrationTest.java
infeo: Fixed in https://github.com/cryptomator/cryptofs/pull/172/commits/2a22af17abc35d271992836146964f9939c25d9b

PR #174, opened 2023-06-06
Use Chunk.data() directly when putting chunk

### LINKED ISSUE #173: Flushing ChunkCache might write incomplete data
When writing to a file channel, for performance reasons cryptofs differs writing on the condition "Does the content to be written fits exaclty in a chunk?"
https://github.com/cryptomator/cryptofs/blob/ec3871dd4556af723fabf37a1be732b1fe576546/src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java#L153-L159

If the condition is true, we directly replace the data in the cache:
https://github.com/cryptomator/cryptofs/blob/ec3871dd4556af723fabf37a1be732b1fe576546/src/main/java/org/cryptomator/cryptofs/fh/ChunkCache.java#L82-L100

Also here a case distinction is made: Is the Chunk not present? On `false`, we overwrite the data of the existing chunk by are using JDKs [`ByteBuffer::duplicate`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/Buffer.html#duplicate()) (getting an independet positon and limit.) and placing the data in the duplicate.

But this approach has the problem, that the chunks data limit is not adjusted! Hence, directly flushing the cache afterwards, this chunk contents will only be saved to the outdated limit before the write, leading to invalid ciphertext.

infeo:
Fixes #173

The chunk data is directly modified. This might lead to data corruption when writing unsynchronized to the same file from different file channels, but the library consumer must handle synchronization.

### DIFF
--- a/src/test/java/org/cryptomator/cryptofs/CryptoFileChannelWriteReadIntegrationTest.java
+++ b/src/test/java/org/cryptomator/cryptofs/CryptoFileChannelWriteReadIntegrationTest.java
@@ -53,7 +53,6 @@ import java.util.Set;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/cryptomator/cryptofs/CryptoFileChannelWriteReadIntegrationTest.java
```diff
@@ -175,6 +174,54 @@ public void afterEach() throws IOException {
...
+					assertRead(reader, 2 * x, 1); //read 1 byte of chunk 1
+					assertWritten(writer, 2 * x, 2 * x); //write to chunk 1
+				}
+			}
+
+			try (var reader2 = FileChannel.open(file, CREATE, READ)) {
+				assertRead(reader2, 2 * x, x); //read chunk 1
+			}
+
+		}
```
overheadhunter, line 193: Is the `read` in between the two writes of any relevance?

Imo the helper functions hurt readability (and aren't used elsewhere in this test):

```suggestion
		@Test
		public void testWritingToPreviouslyReadRegionIsSavedCorrectly() throws IOException {
			int halfChunks = 16_384;
			try (var writer = FileChannel.open(file, CREATE_NEW, WRITE)) {
				writer.writer(ByteBuffer.allocate(3 * halfChunks), 0); // write to chunk 0 and 1
				writer.writer(ByteBuffer.allocate(5 * halfChunks), 0); // write to chunk 0 - 2
			}

			try (var reader = FileChannel.open(file, READ)) {
				// assertAll for reada from chunks 0, 1, 2. 0+2 should be fine even before the fix
			}

		}
```

overheadhunter (commented):
To fully understand the problem/fix, I need more explanation of the "read" in between the writes.

src/test/java/org/cryptomator/cryptofs/CryptoFileChannelWriteReadIntegrationTest.java
infeo: Changed in https://github.com/cryptomator/cryptofs/pull/174/commits/3e6fe77e2b047502a198aabfa535d2fab637985e

overheadhunter (approved):

PR #178, opened 2023-07-29
Add more tests

JaniruTEC:
This PR adds more tests for cryptofs.

PR #179, opened 2023-08-09
Fix: Move directory to existing dir specifiying REPLACE_EXISTING

### LINKED ISSUE #176: DirectoryNotEmptyException when moving directory with REPLACE_EXISTING set
According to the [JavaDocs of Files#move](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Files.html#move(java.nio.file.Path,java.nio.file.Path,java.nio.file.CopyOption...)) moving an empty directory to an existing empty directory should work if `REPLACE_EXISTING` is set: 

> [...] This method may be invoked to move an empty directory. In some implementations a directory has entries for special files or links that are created when the directory is created. In such implementations a directory is considered empty when only the special entries exist. When invoked to move a directory that is not empty then the directory is moved if it does not require moving the entries in the directory. For example, renaming a directory on the same FileStore will usually not require moving the entries in the directory. When moving a directory requires that its entries be moved then this method fails (by throwing an IOException). [...]

> [...] REPLACE_EXISTING If the target file exists, then the target file is replaced if it is not a non-empty directory. If the target file exists and is a symbolic link, then the symbolic link itself, not the target of the link, is replaced. [...]

The following fails with a DirectoryNotEmptyException :
```
var source = fs.getPath("/sourceDir");
var target = fs.getPath("/targetDir");
Files.createDirectory(source);
Files.createDirectory(target);

Assertions.assertDoesNotThrow(() -> Files.move(source, target, REPLACE_EXISTING));
```

Stacktrace:
```
java.nio.file.DirectoryNotEmptyException: /targetDir
	at org.cryptomator.cryptofs/org.cryptomator.cryptofs.CryptoFileSystemImpl.moveDirectory(CryptoFileSystemImpl.java:628)
	at org.cryptomator.cryptofs/org.cryptomator.cryptofs.CryptoFileSystemImpl.move(CryptoFileSystemImpl.java:575)
	at org.cryptomator.cryptofs/org.cryptomator.cryptofs.MoveOperation.move(MoveOperation.java:37)
	at org.cryptomator.cryptofs/org.cryptomator.cryptofs.CryptoFileSystemProvider.move(CryptoFileSystemProvider.java:255)
	at java.base/java.nio.file.Files.move(Files.java:1432)
	[...]
```

The exception is thrown here:
```java
Path oldCiphertextDir = cryptoPathMapper.getCiphertextDir(cleartextTarget).path;
boolean oldCiphertextDirExists = true;
try (DirectoryStream<Path> ds = Files.newDirectoryStream(oldCiphertextDir)) {
	if (ds.iterator().hasNext()) {
		throw new DirectoryNotEmptyException(cleartextTarget.toString()); //628
	}
} catch (NoSuchFileException e) {
	oldCiphertextDirExists = false;
}
```
The offending file is `/vault/d/.../.../dirid.c9r`, presumably the dirid file of the target directory.

Likely resolution: The dirid file of the target directory should not count towards the empty-check.

infeo:
Fixes #176.

Additionally, I changed the execution order when deleting fs nodes of the move target. Now, we first delete the directory link and only then the now orphaned content dir is deleted. The reasoning is: An orphaned directory can be recovered, while a broken directory link cannot be repaired.


src/main/java/org/cryptomator/cryptofs/dir/ExcludeDirIdBackupFilter.java
```diff
@@ -0,0 +1,16 @@
...
+import java.nio.file.Path;
+
+public class ExcludeDirIdBackupFilter implements DirectoryStream.Filter<Path> {
+
+	@Override
+	public boolean accept(Path entry) throws IOException {
+		return !entry.equals(entry.resolveSibling(Constants.DIR_BACKUP_FILE_NAME));
+	}
+
+}
```
overheadhunter, line 16: Can be a constant:

```suggestion
public interface DirListingFilters {

	static DirectoryStream.Filter<Path> EXCLUDE_DIR_ID_BACKUP = p -> !p.equals(p.resolveSibling(Constants.DIR_BACKUP_FILE_NAME));

}
```

src/test/java/org/cryptomator/cryptofs/CryptoFileSystemProviderInMemoryIntegrationTest.java
```diff
@@ -58,6 +58,21 @@ public static void afterAll() throws IOException {
 		tmpFs.close();
 	}
 
+	@Test
+	@DisplayName("Replace an existing, shortened, empty directory")
+	public void testReplaceExistingShortenedDirEmpty() throws IOException {
+		try (var fs = setupCryptoFs(50, 100, false)) {
+			var dirName50Chars = "/target_89_123456789_123456789_123456789_123456789_"; //since filename encryption increases filename length, 50 cleartext chars are sufficient
+			var source = fs.getPath("/sourceDir");
+			var target = fs.getPath(dirName50Chars);
```
overheadhunter, line 67: setting up the fs and source/target can be done `@BeforeEach`, deduplicating some code shared between these two test cases

overheadhunter (changes requested):

src/test/java/org/cryptomator/cryptofs/CryptoFileSystemProviderInMemoryIntegrationTest.java
infeo: It is intentional, that every test set up its own filesystems, not sharing a lot of coce. It would require an additional `@AfterEach` to close the filesystem and also I expect that this class will be populated with more, non-overlapping tests in #178.

src/main/java/org/cryptomator/cryptofs/dir/ExcludeDirIdBackupFilter.java
infeo: Resolved in https://github.com/cryptomator/cryptofs/pull/179/commits/d653636379b166eea02264eb7eeb765b3ae521ff

overheadhunter (approved):

PR #192, opened 2023-12-11
Feature: Refactor OWASP dependency check runs

infeo:
This PRs changes the way, how we use the https://github.com/jeremylong/DependencyCheck maven plugin.

Changes were necessary due to the deprecation of an API used by dependency check (see [here](https://github.com/jeremylong/DependencyCheck#900-upgrade-notice) for details).

Changes:
* Plugin updated to 9.0.4
* plugin uses an NVD API key
* only before a release the plugin is executed
* once a week, the plugin is run on CI. If a vulnerability is detected, the workflow fails and the report is uploaded.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
       interval: "weekly"
       day: "monday"
       time: "06:00"
-      timezone: "UTC"
+      timezone: "Etc/UTC"
     groups:
       java-test-dependencies:
         patterns:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
       interval: "weekly"
       day: "monday"
       time: "06:00"
-      timezone: "Etc/UTC"
+      timezone: "UTC"
     groups:
       java-test-dependencies:
         patterns:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,6 +40,8 @@ jobs:
         with:
           name: artifacts
           path: target/*.jar
+      - uses: ./github/workflows/dependency-check.yml
+        secrets: inherit
       - name: Create release
         if: startsWith(github.ref, 'refs/tags/')
         uses: softprops/action-gh-release@v1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,11 +40,6 @@ jobs:
         with:
           name: artifacts
           path: target/*.jar
-      - name: Check dependencies for releases
-        if: startsWith(github.ref, 'refs/tags/')
-        run: mvn -B verify -Pdependency-check -DskipTests
-        env:
-          NVD_API_KEY: ${{ secrets.NVD_API_KEY }}
       - name: Create release
         if: startsWith(github.ref, 'refs/tags/')
         uses: softprops/action-gh-release@v1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
           restore-keys: ${{ runner.os }}-sonar
       - name: Ensure to use tagged version
         if: startsWith(github.ref, 'refs/tags/')
-        run: mvn versions:set --file ./pom.xml -DnewVersion=${GITHUB_REF##*/}
+        run: mvn -B versions:set --file ./pom.xml -DnewVersion=${GITHUB_REF##*/}
       - name: Build and Test
         run: >
           mvn -B verify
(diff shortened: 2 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -40,6 +40,11 @@ jobs:
         with:
           name: artifacts
           path: target/*.jar
+      - name: Check dependencies for releases
+        if: startsWith(github.ref, 'refs/tags/')
```
overheadhunter, line 44: Don't run on tagged commits, but rather on release-branches. Failing here might be too late, as in some repos a lot of automation happens when pushing tags.

.github/dependabot.yml
```diff
@@ -6,7 +6,7 @@ updates:
       interval: "weekly"
       day: "monday"
       time: "06:00"
-      timezone: "UTC"
+      timezone: "Etc/UTC"
```
overheadhunter, line 9: unrelated

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          show-progress: false
+      - name: Setup Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: 17
+          cache: 'maven'
+      - name: Run org.owasp:dependency-check plugin
+        id: plugin-run
```
overheadhunter, line 23: choose a more descriptive name, e.g.

```suggestion
        id: dependency-check
```

overheadhunter (changes requested):
good start, but how are we notified? just a failed workflow? that might be a lot of spam for the repo owner. instead, can we post such notifications in slack?

infeo:
> good start, but how are we notified? just a failed workflow? that might be a lot of spam for the repo owner. instead, can we post such notifications in slack?

I had exactly the same thoughts with the reversed result 😆  But will change it to a slack notification.

.github/workflows/dependency-check.yml
infeo: Changed in https://github.com/cryptomator/cryptofs/pull/192/commits/60b98e43a14a8b3d7e2bb835307cbef05c4ecf4b

.github/dependabot.yml
infeo: Reverted in https://github.com/cryptomator/cryptofs/pull/192/commits/60b98e43a14a8b3d7e2bb835307cbef05c4ecf4b

.github/workflows/build.yml
infeo: Changed in https://github.com/cryptomator/cryptofs/pull/192/commits/60b98e43a14a8b3d7e2bb835307cbef05c4ecf4b

infeo:
> notifications in slack

Done in https://github.com/cryptomator/cryptofs/pull/192/commits/8e20b707f52e62dd33db206ed6bf827c9c06df15.

@overheadhunter The only part missing is now the merge constraint. Can you add it to the repository? It should be very similar to https://github.com/cryptomator/cryptomator and its `release-check.yml` action.

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,55 @@
...
+          NVD_API_KEY: ${{ secrets.NVD_API_KEY }}
+      - name: Upload report on failure
+        if: "${{ steps.dependency-check.outcome == 'failure'}}"
+        uses: actions/upload-artifact@v3
+        with:
+          name: dependency-check-report
+          path: target/dependency-check-report.html
+          if-no-files-found: error
+      - name: Slack Notification on regular check
+        if: github.event_name == 'schedule'
```
overheadhunter, line 40: Doesn't this need some `&& foo == failure`? Or does this only apply to the first consecutive step after running the depencency check?

overheadhunter:
> The only part missing is now the merge constraint. Can you add it to the repository?

I can't reference the action from the repo settings, before it exists on the default branch

.github/workflows/dependency-check.yml
```diff
@@ -0,0 +1,54 @@
...
+          java-version: 17
+          cache: 'maven'
+      - name: Run org.owasp:dependency-check plugin
+        id: dependency-check
+        continue-on-error: true
+        run: mvn -B verify -Pdependency-check -DskipTests
+        env:
+          NVD_API_KEY: ${{ secrets.NVD_API_KEY }}
+      - name: Upload report on failure
+        if: "${{ steps.dependency-check.outcome == 'failure'}}"
```
overheadhunter, line 32: can't this be just

```suggestion
        if: steps.dependency-check.outcome == 'failure'
```

overheadhunter (changes requested):

overheadhunter (approved):

.github/workflows/dependency-check.yml
infeo: Done in https://github.com/cryptomator/cryptofs/pull/192/commits/d523d09817feef33e03e86c58a0c5a696481543b

PR #206, opened 2024-02-02
Flush ciphertext channel before writing lastModified timestamp

### LINKED ISSUE #205: Ciphertext filechannels close() might change lastModified timestamp
When opening a filechannel to a not previously opened file in cryptofs, an `OpenCryptoFile` for this file (path) is instantiated and `OpenCryptoFile::newFileChannel`-method is called on it.

There the ciphertext filechannel (which is the one actually writing to the underlying filesystem) is created and linked to the returned cleartext file channel. When closing the cleartext file channel via `channel.close()`, the "close chain" runs over JDKs `AbstractInteruptibleChannel.close()` and the implementations of the `implCloseChannel()` methods in `AbstractFileChannel` and last and most importantly, the `CleartextFileChannel`:
https://github.com/cryptomator/cryptofs/blob/6f04f2a2677ca4104d2d34e867aebaa35d877048/src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java#L322-L337

* The `flush()`-method writes, if necessary the file header and otherwise empties the chunk cache.
* Afterwards, the `lastModified` timestamp is corrected (might be wrong on storage due to caching). 
* Finally, the super impl of `implCloseChannel` (does not perform any IO)
* and a close listener is called. The latter is a method in `OpenCryptoFile`, set via the Dagger `ChannelComponent`:
https://github.com/cryptomator/cryptofs/blob/6f04f2a2677ca4104d2d34e867aebaa35d877048/src/main/java/org/cryptomator/cryptofs/fh/OpenCryptoFile.java#L186-L198

It basically deregisters the channel from some pools/queues and closes the ciphertext file channel.

Even though we explicitly write the last modifed time to storage, it might be changed afterwards because we never flush the ciphertextchannel explicitly before setting the last modified time. Hence, if there are unwritten changes in the channel, it is only written _on close_, which is after correcting the last modified time, effectively nullifying the correction.

The `flush` call only regards the chunk cache, which might use other (ciphertext) file channels.

infeo:
Fixes #205.

Unfortunately, the changes became bigger than anticipated. Reason is, that due to our ..."spaghetti" architecture, with the minimal fix could not exclude the possibility that the lastModified timestamp is not overwritten. The chunkCache uses an arbitrary ciphertext filechannel to write to a file via the `ChunkIO` class. It was theoretically possible to first get a writable channel but deregister this channel in a different thread. With my implementation one can either deregister channels to the same file or write into such channels, but cannot do both. Deregistering is prioritized to timely finish close actions.

With this PR we have now the guarantee, that between a ciphertext channels `force(...)` and `close()` method, no other thread can write to it, hence persisting the lastModified timestamp can only fail on the storage side. Additionally, i moved all IO-related things during close to the `CleartextFileChannel` (some were lingering around in `OpenCryptoFile`) and improved the unit tests.

## Summary by CodeRabbit

- **New Features**
	- Introduced a more sophisticated synchronization mechanism for file operations, enhancing thread safety and performance.
- **Refactor**
	- Updated file channel handling to improve reliability and efficiency in file operations.
	- Simplified file channel interaction for better maintainability.
- **Bug Fixes**
	- Adjusted file channel closure handling to ensure proper resource management.
- **Tests**
	- Enhanced testing for file channel operations, including handling of `IOExceptions`.
	- Improved test accuracy by refining mock object creation and method calls.

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/fh/ChunkIO.java
+++ b/src/main/java/org/cryptomator/cryptofs/fh/ChunkIO.java
@@ -6,7 +6,6 @@ import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
 import java.nio.channels.NonReadableChannelException;
 import java.nio.channels.NonWritableChannelException;
-import java.util.Iterator;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -15,6 +14,8 @@ class ChunkIO {
 
 	private final Set<FileChannel> readableChannels = ConcurrentHashMap.newKeySet();
 	private final Set<FileChannel> writableChannels = ConcurrentHashMap.newKeySet();
+
+	private final PriorityMutex synchronizer = new PriorityMutex();
 
 	@Inject
 	public ChunkIO() {
@@ -39,8 +40,10 @@ class ChunkIO {
 	 * @param channel
 	 */
 	public void unregisterChannel(FileChannel channel) {
-		readableChannels.remove(channel);
-		writableChannels.remove(channel);
+		try (var token = synchronizer.dispensePriority()) {
+			readableChannels.remove(channel);
+			writableChannels.remove(channel);
+		}
 	}
 
 	int read(ByteBuffer dst, long position) throws IOException {
@@ -48,25 +51,17 @@ class ChunkIO {
 	}
 
 	int write(ByteBuffer src, long position) throws IOException {
-		return getWritableChannel().write(src, position);
+		try (var token = synchronizer.dispenseRegular()) {
+			return getWritableChannel().write(src, position);
+		}
 	}
 
 	private FileChannel getReadableChannel() {
-		Iterator<FileChannel> iter = readableChannels.iterator();
-		if (iter.hasNext()) {
-			return iter.next();
-		} else {
-			throw new NonReadableChannelException();
-		}
+		return readableChannels.stream().findAny().orElseThrow(NonReadableChannelException::new);
 	}
 
 	private FileChannel getWritableChannel() {
-		Iterator<FileChannel> iter = writableChannels.iterator();
-		if (iter.hasNext()) {
-			return iter.next();
-		} else {
-			throw new NonWritableChannelException();
-		}
+		return writableChannels.stream().findAny().orElseThrow(NonWritableChannelException::new);
 	}
 
 }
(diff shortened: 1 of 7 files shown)


src/test/java/org/cryptomator/cryptofs/ch/CleartextFileChannelTest.java
```diff
@@ -69,7 +68,7 @@ public class CleartextFileChannelTest {
 	private FileHeaderHolder headerHolder = mock(FileHeaderHolder.class);
 	private AtomicBoolean headerIsPersisted = mock(AtomicBoolean.class);
 	private EffectiveOpenOptions options = mock(EffectiveOpenOptions.class);
-	private Path filePath = Mockito.mock(Path.class,"/foo/bar");
+	private Path filePath = Mockito.mock(Path.class, "/foo/bar");
```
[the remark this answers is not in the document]
overheadhunter: The argument in question is the mock's name. In this case, the mock is intended to be named like a file path, as it represents a certain path. This is purely for readability during debug sessions, as this name is returned by `toString()`.

overheadhunter (changes requested):
Tbh, this is too big of a change to easily review it without a deep understanding of what is going on here.

Your PR description suggests there is some race condition between closing and writing. I suggest to create a new bug report (including a sequence diagram to explain this!) and solve this in a separate PR *first*. Then come back to fixing the last modified date.

Furthermore I suspect changing the order of `persistLastModified()` and calling the closeListener in `CleartextFileChannel#implCloseChannel` might cause regressions. Besides, I can not approve this without fully understanding it and solving complexity by adding more complexity is the very cause of this lib becoming less and less maintainable.

src/main/java/org/cryptomator/cryptofs/fh/ChunkIO.java
```diff
@@ -16,6 +15,8 @@ class ChunkIO {
 	private final Set<FileChannel> readableChannels = ConcurrentHashMap.newKeySet();
 	private final Set<FileChannel> writableChannels = ConcurrentHashMap.newKeySet();
 
+	private final PriorityMutex synchronizer = new PriorityMutex();
```
overheadhunter, line 18: A conventional ReadWriteLock should suffice:

* readlock: allow multiple threads to enter `write()`
* writelock: only allow a thread to enter `unregisterChannel()` when nobody is `write()`ing

overheadhunter: which has the advantage that there is no need to write plenty of tests covering concurrency issues with a custom synchronization means.

infeo:
I'll open a new PR with fewer changes.

PR #228, opened 2024-06-20
Feature: First close file channel, then persist lastModified

infeo:
This PR changes the logic of persisting the lastModfied date from _before_ closing the ciphertext channel to _afterwards_. Additionally, the cryptofs internal lastModifiedDate is always updated, even when persisting to storage fails.

This impl change is motivated by the behaviour of some fs/storage implemenations, which set the lastModified time to "now" when closing a channel, regardless of changes made.

Before merging, we should test the implementation with different scenarios:
|         | OneDrive | pCloud | GoogleDrive | iCloud | Dropbox | LocalFS |
|---------|----------|--------|-------------|--------|----------|----------|
| Windows |      ✅    |  ✅    |     ✅         |   ❓1      |      ✅    |     ✅    |
| macOS   |     ✅     |   ✅     |     ✅        |   ✅     |  ✅ | ✅  | 
| Linux   |    ➖ 2      |   ➖ 2     |    ➖ 2      |   ➖ 2   | ➖ 2   |  ✅  |

The test was to edit and save an spreadsheet/document file and copying files of different sizes into the vault.

1: iCloud on Windows is... strange. First of all, everything copied into the iCloud dir gets as mTime the _current_ time. Inside a Cryptomator vault, this becomes even stranger. Sometimes, the files are set to the current time, but especially bigger files (<30MB) get ... "some" time. "Some" can be the current, the former file mTime or something in between.

2: On Linux there are no offical filesystem integrations. GNOME offers "online account", but these are not accessible from the filesystem.

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/attr/CryptoBasicFileAttributeView.java
+++ b/src/main/java/org/cryptomator/cryptofs/attr/CryptoBasicFileAttributeView.java
@@ -48,10 +48,10 @@ sealed class CryptoBasicFileAttributeView extends AbstractCryptoFileAttributeVie
 	@Override
 	public void setTimes(FileTime lastModifiedTime, FileTime lastAccessTime, FileTime createTime) throws IOException {
 		readonlyFlag.assertWritable();
-		getCiphertextAttributeView(BasicFileAttributeView.class).setTimes(lastModifiedTime, lastAccessTime, createTime);
 		if (lastModifiedTime != null) {
 			getOpenCryptoFile().ifPresent(file -> file.setLastModifiedTime(lastModifiedTime));
 		}
+		getCiphertextAttributeView(BasicFileAttributeView.class).setTimes(lastModifiedTime, lastAccessTime, createTime);
 	}
 
 }
--- a/src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java
+++ b/src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java
@@ -326,19 +326,29 @@ public class CleartextFileChannel extends AbstractFileChannel {
 		try {
 			flush();
 		} finally {
+			implCloseChannelFinally1();
+		}
+	}
+
+	private void implCloseChannelFinally1() throws IOException {
+		try {
 			super.implCloseChannel();
+		} finally {
 			closeListener.closed(this);
+			implCloseChannelFinally2();
+		}
+	}
+
+	private void implCloseChannelFinally2() throws IOException {
+		try {
+			ciphertextFileChannel.close();
+		} finally {
 			try {
-				ciphertextFileChannel.close();
-			} finally {
-				try {
-					persistLastModified();
-				} catch (NoSuchFileException nsfe) {
-					//no-op, see https://github.com/cryptomator/cryptofs/issues/169
-				} catch (IOException e) {
-					//only best effort attempt
-					LOG.warn("Failed to persist last modified timestamp for encrypted file: {}", e.getMessage());
-				}
+				persistLastModified();
+			} catch (NoSuchFileException nsfe) {
+				//no-op, see https://github.com/cryptomator/cryptofs/issues/169
+			} catch (IOException e) {
+				LOG.warn("Failed to persist last modified timestamp for encrypted file: {}", e.getMessage());
 			}
 		}
 	}
--- a/src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java
+++ b/src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java
@@ -322,6 +322,10 @@ public class CleartextFileChannel extends AbstractFileChannel {
 	protected void implCloseChannel() throws IOException {
 		try {
 			flush();
+		} finally {
+			super.implCloseChannel();
+			closeListener.closed(this);
+			ciphertextFileChannel.close();
 			try {
 				persistLastModified();
 			} catch (NoSuchFileException nsfe) {
@@ -330,9 +334,6 @@ public class CleartextFileChannel extends AbstractFileChannel {
 				//only best effort attempt
 				LOG.warn("Failed to persist last modified timestamp for encrypted file: {}", e.getMessage());
 			}
-		} finally {
-			super.implCloseChannel();
-			closeListener.closed(this);
 		}
 	}
 }
--- a/src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java
+++ b/src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java
@@ -1,5 +1,6 @@
 package org.cryptomator.cryptofs.ch;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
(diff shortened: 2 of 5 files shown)


src/main/java/org/cryptomator/cryptofs/attr/CryptoBasicFileAttributeView.java
```diff
@@ -48,10 +48,10 @@ public BasicFileAttributes readAttributes() throws IOException {
 	@Override
 	public void setTimes(FileTime lastModifiedTime, FileTime lastAccessTime, FileTime createTime) throws IOException {
 		readonlyFlag.assertWritable();
-		getCiphertextAttributeView(BasicFileAttributeView.class).setTimes(lastModifiedTime, lastAccessTime, createTime);
 		if (lastModifiedTime != null) {
 			getOpenCryptoFile().ifPresent(file -> file.setLastModifiedTime(lastModifiedTime));
 		}
+		getCiphertextAttributeView(BasicFileAttributeView.class).setTimes(lastModifiedTime, lastAccessTime, createTime);
```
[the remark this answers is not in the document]
infeo: The error handling is delegated to the callee.

src/test/java/org/cryptomator/cryptofs/ch/CleartextFileChannelTest.java
```diff
@@ -236,20 +235,26 @@ public void testForceWithoutMetadataDoesntUpdatesLastModifiedTime() throws IOExc
...
+		public void testCloseIoExceptionFlush() throws IOException {
+			var inSpy = Mockito.spy(inTest);
+			Mockito.doThrow(IOException.class).when(inSpy).flush();
+
+			Assertions.assertThrows(IOException.class, () -> inSpy.implCloseChannel());
 
-			verify(closeListener).closed(inTest);
+			verify(closeListener).closed(inSpy);
+			verify(ciphertextFileChannel).close();
+			verify(inSpy).persistLastModified();
```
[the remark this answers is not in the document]
infeo: Since this is a unit test, we do not have to handle the exceptions. If the exception is thrown, the test fails.

src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java
```diff
@@ -324,18 +325,21 @@ long beginOfChunk(long cleartextPos) {
...
+				} catch (NoSuchFileException nsfe) {
+					//no-op, see https://github.com/cryptomator/cryptofs/issues/169
+				} catch (IOException e) {
+					//only best effort attempt
+					LOG.warn("Failed to persist last modified timestamp for encrypted file: {}", e.getMessage());
+				}
+			}
 		}
 	}
 }
```
overheadhunter, line 345: as per the (badly designed) nio API, `implCloseChannel` is called at most once, regardless of whether closing succeeded or failed. I.e. *everything* we do here must be "best effort" style, there is no chance to re-attempt anything after recovering from an exception.

In other words, all ~four~ five statements must be invoked: `flush()`, `super.implCloseChannel()`, `closeListener.closed(this)`, `ciphertextFileChannel.close()` (+ bonus `persistLastModified()`).

As any of these can throw (e.g. `closeListener.closed(this)`), I believe there is no way around an absurdly nested `try-finally` monster. To keep this readable, I suppose we need to split this up into multiple methods.

src/main/java/org/cryptomator/cryptofs/fh/OpenCryptoFile.java
```diff
@@ -183,17 +183,13 @@ public void updateCurrentFilePath(Path newFilePath) {
...
-			if (openChannels.isEmpty()) {
-				close();
-			}
+	private synchronized void channelClosed(CleartextFileChannel cleartextFileChannel) {
+		FileChannel ciphertextFileChannel = openChannels.remove(cleartextFileChannel);
+		if (ciphertextFileChannel != null) {
+			chunkIO.unregisterChannel(ciphertextFileChannel);
+		}
+		if (openChannels.isEmpty()) {
+			close();
```
overheadhunter, line 192: Since closing the ciphertext file channel moved to the CleartextFileChannel, we should change the type of `openChannels` to a Set (use `ConcurrentHashMap.newKeySet()`)

overheadhunter (changes requested):

src/main/java/org/cryptomator/cryptofs/fh/OpenCryptoFile.java
infeo: We cannot change the type of `openChannels` at the moment, because we still need the ciphertextFileChannel for removing it from the chunkIO class.

If we inject the chunkIO object to the cleartextFileChannel, we could refactor it. I would also favor this, but it is out of scope of this PR.

src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java
infeo: i split the method into three parts in e247a2e264eb0a6f9e2cff53e77238678d43b842

overheadhunter: I dislike the names but don't have any better suggestions.

Maybe use a helper function?

```java
protected void implCloseChannel() throws IOException {
	tryAll(List.<CloseAction>of(
		this::flush,
		super::implCloseChannel,
		ciphertextFileChannel::close,
		this::tryPersistLastModified
	).iterator());
}

private void tryAll(Iterator<CloseAction> actions) throws IOException {
	if (actions.hasNext()) {
		try {
			actions.next().run();
		} finally {
			tryAll(actions);
		}
	}
}

@FunctionalInterface
private interface CloseAction {
	void run() throws IOException;
}
```

infeo: Oh, i like this idea. Keeps the important functions in one place and readable. Changed in 675780129dfacd5cb6f8f0ebcd6ad9b13942eee0

src/main/java/org/cryptomator/cryptofs/ch/CleartextFileChannel.java
```diff
@@ -322,20 +325,38 @@ long beginOfChunk(long cleartextPos) {
...
-			closeListener.closed(this);
 		}
 	}
+
+	@FunctionalInterface
+	private interface CloseAction {
+
+		void run() throws IOException;
+	}
+
```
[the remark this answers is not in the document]
overheadhunter: No risk as long as the list of actions is hard-coded. The iterative approach is fine as well, but this specific suggestion swallows exceptions.

Can you update your proposal to re-throw the first caught exception after the loop and adding further exceptions as suppressed exceptions to the first one, if necessary?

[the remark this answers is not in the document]
overheadhunter: @infeo feel free to decide which approach you like best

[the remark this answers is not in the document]
infeo: I keep the recursive approach.

overheadhunter (approved):

PR #270, opened 2025-01-10
Use truncate() of CleartextChannel when opening new Channel with `TRUNCATE_EXISTING`

### LINKED ISSUE #269: Truncating a channel which is currently written can create invalid ciphertext
## Summary
Library version: 2.7.1

Truncating a channel which is currently written to can create invalid ciphertext/unreadable file.

## Description
In our test suite exists a test, which truncates a file while another thread writes data to the same file:
https://github.com/cryptomator/cryptofs/blob/4d6a56b9c221511a4b6910bd44b5e023321e82e2/src/test/java/org/cryptomator/cryptofs/CryptoFileChannelWriteReadIntegrationTest.java#L620-L647

This test fails in very rare cases, e.g. in CI run [12694328432](https://github.com/cryptomator/cryptofs/actions/runs/12694328432). If the logs are not available anymore, you can download them [here](https://github.com/user-attachments/files/18380545/0_Build.and.Test.txt)

You can trigger the failure with higher probability, if you add a new line after line 634 with `Thread.sleep(2000);`.

infeo:
This PR fixes #269.

Instead of writing an "optimized", but wrong truncate block, we can simply use `CleartextChannel.truncate(0)` when opening a new file. Benefit of this method: It is already tested! And since it is synchronized and also creating a new file channel is synchronized, we can be sure that side calls cannot interfere.

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/fh/OpenCryptoFile.java
+++ b/src/main/java/org/cryptomator/cryptofs/fh/OpenCryptoFile.java
@@ -70,15 +70,13 @@ public class OpenCryptoFile implements Closeable {
 		try {
 			ciphertextFileChannel = path.getFileSystem().provider().newFileChannel(path, options.createOpenOptionsForEncryptedFile(), attrs);
 			initFileHeader(options, ciphertextFileChannel);
-			if (options.truncateExisting()) {
-				chunkCache.invalidateStale();
-				ciphertextFileChannel.truncate(cryptor.fileHeaderCryptor().headerSize());
-				fileSize.set(0);
-			}
 			initFileSize(ciphertextFileChannel);
 			cleartextFileChannel = component.newChannelComponent() //
 					.create(ciphertextFileChannel, options, this::cleartextChannelClosed) //
 					.channel();
+			if (options.truncateExisting()) {
+				cleartextFileChannel.truncate(0);
+			}
 		} finally {
 			if (cleartextFileChannel == null) { // i.e. something didn't work
 				cleartextChannelClosed(ciphertextFileChannel);
--- a/src/main/java/org/cryptomator/cryptofs/fh/OpenCryptoFile.java
+++ b/src/main/java/org/cryptomator/cryptofs/fh/OpenCryptoFile.java
@@ -26,7 +26,6 @@ public class OpenCryptoFile implements Closeable {
 
 	private final FileCloseListener listener;
 	private final AtomicReference<Instant> lastModified;
-	private final ChunkCache chunkCache;
 	private final Cryptor cryptor;
 	private final FileHeaderHolder headerHolder;
 	private final ChunkIO chunkIO;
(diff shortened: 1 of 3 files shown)


src/test/java/org/cryptomator/cryptofs/CryptoFileChannelWriteReadIntegrationTest.java
```diff
@@ -632,6 +632,7 @@ public void testConcurrentWriteAndTruncate() throws IOException {
 						buf.flip();
 					}
 				});
+				Thread.sleep(1000);
```
overheadhunter, line 635: To be removed for CI, rather use a barrier, countdownlatch, etc.

src/test/java/org/cryptomator/cryptofs/fh/OpenCryptoFileTest.java
```diff
@@ -93,19 +93,20 @@ public void testCloseImmediatelyIfOpeningFirstChannelFails() {
 	}
 
 	@Test
-	@DisplayName("Opening a file channel with TRUNCATE_EXISTING sets the file size to 0")
+	@DisplayName("Opening a file channel with TRUNCATE_EXISTING calls truncate(0) on the cleartextChannel")
 	public void testFileSizeZerodOnTruncateExisting() throws IOException {
```
overheadhunter, line 97: Method name misleading, since you removed the file size verification.

I assume file size is implicitly zeroed in the truncate call?

src/main/java/org/cryptomator/cryptofs/fh/OpenCryptoFile.java
```diff
@@ -70,15 +70,13 @@ public synchronized FileChannel newFileChannel(EffectiveOpenOptions options, Fil
 		try {
 			ciphertextFileChannel = path.getFileSystem().provider().newFileChannel(path, options.createOpenOptionsForEncryptedFile(), attrs);
 			initFileHeader(options, ciphertextFileChannel);
-			if (options.truncateExisting()) {
-				chunkCache.invalidateStale();
-				ciphertextFileChannel.truncate(cryptor.fileHeaderCryptor().headerSize());
-				fileSize.set(0);
-			}
 			initFileSize(ciphertextFileChannel);
```
overheadhunter, line 73: can be removed/replaced as well, can't it? The old ciphertextFileSize is most likely wrong at this point in time.

overheadhunter (changes requested):
Good catch, did you spot it in the failed CI run?

infeo:
> did you spot it in the failed CI run?

yes. I worked on a completely different feature (FileNameDecryptor) and was tempted to declare the failing build as a fluke. Luckily i decided to investigate the issue^^ We have tests for a reason!

src/main/java/org/cryptomator/cryptofs/fh/OpenCryptoFile.java
infeo: It could be replaced, but it is not related to the issue.
The method initializes the file size variable on any _newly_ opened file. If the filesize is already initatilizes, the method directly returns.

So, if we opened the first time a file and truncate it, it does not matter because we are inside a synchronized block and no other thread can read wrong information.
If it is not the first time (aka fileSize is already initialized), the method does nothing, which is also okay.

src/test/java/org/cryptomator/cryptofs/fh/OpenCryptoFileTest.java
infeo: > I assume file size is implicitly zeroed in the truncate call?

Yes, but this is not the concern of this unit test any more. I renamed the method in https://github.com/cryptomator/cryptofs/pull/270/commits/38135c935e5da7c5bd55dfd6ea3ef70425257865

src/test/java/org/cryptomator/cryptofs/CryptoFileChannelWriteReadIntegrationTest.java
infeo: Done in https://github.com/cryptomator/cryptofs/pull/270/commits/9d93701d58d827aeb5dfbe34148e8e75cfdf37b6

overheadhunter (approved):

src/test/java/org/cryptomator/cryptofs/CryptoFileChannelWriteReadIntegrationTest.java
overheadhunter: for the record, I meant like in 5363d4b

Issue #89, opened 2020-08-20
Add API point to add an event listener


infeo:
## Summary
Add an API to hand over/add an event listener to cryptofs. The event listener is bidirectional, i.e. does not only report but also returns a value. It is our decision to which events the listener is triggered (e.g. failed reading of a file, not existing directory, etc)

## Problem Description
In cryptofs there are cases where it would be good to inform the caller about certain events/findings (e.g. files not belonging to the vault structure, invalid dir directories). Currently, only an error is thrown and the operation aborted or the event is logged, but otherwise ignored. From the caller perspective this is not the most desireable case, because with this you might miss (for you) important findings.

## Solution
Expose in the API a method to add an eventlistener and then notify the listener on predefined events. The predefined events should be documented somewhere. For example there could be an extended method for filesystem creation, where one can hand over the event listener.

## Alternatives
Collecting all occuring, not critical exceptions and then hand them over (in some way) to the caller. But this would require again some side channel.

PR #277, opened 2025-02-06
Feature: Notification API for Filesystem events

infeo:
This PR adds a notification API to the filesytem and closes https://github.com/cryptomator/cryptofs/issues/89.

The notification API is simple: At filesystem creation the caller hands over a Consumer<FilesystemEvent> object via the CryptoFilesystemProperties map. Everytime, the filesystem creates an event, it is consumed by this object.

For the start there are 3 different events:
* DecryptionFailedEvent
* ConflictResolvedEvent
* ConflictResolutionFailedEvent

Supersedes #276 by reducing the diff.

Remark: The DecryptionFailed event will be enhanced with the cleartext path in a different PR

### DIFF
--- a/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemProperties.java
+++ b/src/main/java/org/cryptomator/cryptofs/CryptoFileSystemProperties.java
@@ -86,7 +86,7 @@ public class CryptoFileSystemProperties extends AbstractMap<String, Object> {
 	 *
 	 * @since 2.9.0
 	 */
-	public static final String PROPERTY_EVENT_CONSUMER = "notificationConsumer";
+	public static final String PROPERTY_EVENT_CONSUMER = "fsEventConsumer";
 
 	static final Consumer<FilesystemEvent> DEFAULT_EVENT_CONSUMER = ignored -> {};
 


src/main/java/org/cryptomator/cryptofs/CryptoFileSystemProperties.java
```diff
@@ -80,6 +81,15 @@ public class CryptoFileSystemProperties extends AbstractMap<String, Object> {
...
 	static final String DEFAULT_MASTERKEY_FILENAME = "masterkey.cryptomator";
 
+	/**
+	 * Key identifying the function to call for notifications.
+	 *
+	 * @since 2.9.0
+	 */
+	public static final String PROPERTY_NOTIFY_METHOD = "notificationConsumer";
+
+	static final Consumer<FilesystemEvent> DEFAULT_NOTIFY_METHOD = (FilesystemEvent e) -> {};
```
overheadhunter, line 91: To keep naming coherent, I'd suggest:

```suggestion
	public static final String PROPERTY_EVENT_CONSUMER = "fsEventConsumer";
	
	static final Consumer<FilesystemEvent> DEFAULT_EVENT_CONSUMER = (FilesystemEvent e) -> {};

```

src/main/java/org/cryptomator/cryptofs/dir/C9rConflictResolver.java
```diff
@@ -111,6 +121,7 @@ private Node renameConflictingFile(Path canonicalPath, Path conflictingPath, Str
 		Node node = new Node(alternativePath);
 		node.cleartextName = alternativeCleartext;
 		node.extractedCiphertext = alternativeCiphertext;
+		eventConsumer.accept(new ConflictResolvedEvent(cleartextPath.resolve(cleartext), canonicalPath, cleartextPath.resolve(alternativeCleartext), alternativePath));
```
overheadhunter, line 124: third param `cleartextPath.resolve(alternativeCleartext)` is either not what the event object wants here or all the different properties are pretty confusing.

> oldVersionCleartextPath path within the cryptographic filesystem of the renamed resource

Is there really such a thing as an "old cleartext"? And is there any benefit of displaying this to the user? I'd drop this.

src/main/java/org/cryptomator/cryptofs/event/ConflictResolvedEvent.java
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Emitted, if a conflict inside an encrypted directory was resolved
+ *
+ * @param cleartextPath path within the cryptographic filesystem
+ * @param ciphertextPath path to the encrypted resource
+ * @param oldVersionCleartextPath path within the cryptographic filesystem of the renamed resource
+ * @param oldVersionCiphertextPath path to the renamed, encrypted resource
+ */
+public record ConflictResolvedEvent(Path cleartextPath, Path ciphertextPath, Path oldVersionCleartextPath, Path oldVersionCiphertextPath) implements FilesystemEvent {
```
overheadhunter, line 13: as said before, I'd just use `oldCiphertextPath`, `newCiphertextPath` and `cleartextPath`

overheadhunter: also: Should this the canonical cleartext path? Or the on of the renamed resource?

src/main/java/org/cryptomator/cryptofs/event/FilesystemEvent.java
```diff
@@ -0,0 +1,27 @@
...
+ * Common interface for all filesystem events.
+ * <p>
+ * Events are emitted via the notification method set in the properties during filesystem creation, see {@link org.cryptomator.cryptofs.CryptoFileSystemProperties.Builder#withFilesystemEventConsumer(Consumer)}.
+ * <p>
+ * To get a specific event type, use the enhanced switch pattern or typecasting in if-instance of, e.g.
+ * {@code
+ * FilesystemEvent fse;
+ * switch (fse) {
+ *   case DecryptionFailedEvent dfe -> //do stuff
+ *   case ConflictResolvedEvent cre -> //do other stuff
```
overheadhunter, line 15: don't encourage people to use weird abbreviations for variables that don't even share the same scope

```suggestion
 *   case DecryptionFailedEvent e -> //do stuff
 *   case ConflictResolvedEvent e -> //do other stuff
```

src/main/java/org/cryptomator/cryptofs/fh/FileHeaderHolder.java
```diff
@@ -75,6 +81,9 @@ FileHeader loadExisting(FileChannel ch) throws IOException {
 			isPersisted.set(true);
 			return existingHeader;
 		} catch (IllegalArgumentException | CryptoException e) {
+			if (e instanceof AuthenticationFailedException afe) {
+				eventConsumer.accept(new DecryptionFailedEvent(path.get(), afe));
+			}
```
overheadhunter, line 86: As mentioned above, I'd drop this type check. The user probably doesn't care for the technical reason causing the decryption to fail.

overheadhunter (changes requested):

src/main/java/org/cryptomator/cryptofs/event/DecryptionFailedEvent.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.nio.file.Path;
+
+/**
+ * Emitted, if a decryption operation fails.
+ *
+ * @param ciphertextPath path to the encrypted resource
+ * @param e thrown exception
+ */
+public record DecryptionFailedEvent(Path ciphertextPath, AuthenticationFailedException e) implements FilesystemEvent {
```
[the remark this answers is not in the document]
infeo: I intentionally only catched AuthFailed exceptions: The event notifies the user about invalid encrypted content. Any IO related error is "normal" filesystem behaviour and might be fixed elsewhere by a sync, while the latter cannot be normally fixed.

src/main/java/org/cryptomator/cryptofs/dir/C9rConflictResolver.java
infeo: old is from the point of syncing 😅 I'll rename the parameters

The event should give the library consumer as much information as possible. Therefore, i added 4 parameters:
* the old, conflicting ciphertext name
* the cleartext name of the conflicting resource
* the new, non-conlict ciphertext name
* the new cleartext name

Especially with our new rename algorithm in #275 we should keep the cleartextname of the the conflicting resource.

infeo: Changed in https://github.com/cryptomator/cryptofs/pull/277/commits/2cd07fae1fc779bfa798ba287919ba9cb69227af.

src/main/java/org/cryptomator/cryptofs/event/ConflictResolvedEvent.java
infeo: Changed in https://github.com/cryptomator/cryptofs/pull/277/commits/2cd07fae1fc779bfa798ba287919ba9cb69227af

src/main/java/org/cryptomator/cryptofs/CryptoFileSystemProperties.java
```diff
@@ -153,6 +164,10 @@ int shorteningThreshold() {
 		return (int) get(PROPERTY_SHORTENING_THRESHOLD);
 	}
 
+	Consumer<FilesystemEvent> fsEventConsumner() {
+		return (Consumer<FilesystemEvent>) get(PROPERTY_NOTIFY_METHOD);
+	}
```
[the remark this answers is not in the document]
infeo: fixed in https://github.com/cryptomator/cryptofs/pull/277/commits/fb5b30207d11f921d2eae57c4102d3f81b2abbb7

src/main/java/org/cryptomator/cryptofs/CryptoFileSystemProperties.java
infeo: Renamed in https://github.com/cryptomator/cryptofs/pull/277/commits/3691c6bd5a17901ffee920e83e3bae7e3ce70d06

src/main/java/org/cryptomator/cryptofs/CryptoFileSystemProperties.java
```diff
@@ -80,6 +81,15 @@ public class CryptoFileSystemProperties extends AbstractMap<String, Object> {
 
 	static final String DEFAULT_MASTERKEY_FILENAME = "masterkey.cryptomator";
 
+	/**
+	 * Key identifying the function to call for notifications.
+	 *
+	 * @since 2.9.0
+	 */
+	public static final String PROPERTY_EVENT_CONSUMER = "notificationConsumer";
```
overheadhunter, line 89: Change the value as well, since being used as a map key it is essentially part of the public API

```suggestion
	public static final String PROPERTY_EVENT_CONSUMER = "fsEventConsumer";
```

overheadhunter (changes requested):

overheadhunter (approved):

PR #278, opened 2025-02-11
Feature: Add cleartext path to decryption failed event

infeo:
This PR improves the decryption failed event by adding the cleartext path to it, making the event more meaningful.

Unfortunately, a lot of adjustments had to be made to make it possible. The problem was that inside the file handle package (containing `OpenCryptoFile`, `ChunkCache` and other decrytion related classes) the cleartext path did not exist anymore since everything could be done solely working with the ciphertext path.

With this PR, a new record class is introduced: `ClearAndCipherPath`. It represents a resource inside the cryptographic file system (symlink, directory or file) and stores its cleartext name (visible by the consumer) and its (principal) ciphertext path on the harddisk. Since the invariant "the cleartext path changes if and only if the ciphertext path changes` holds, it is quite logical to group those in a single object and update them simultanously.

Coming back to the goal of the PR, the `ClearAndCipherPath`  object is handed over to the `OpenCryptoFileModule` class, such that it is available in the `fh` package and the cleartext path is available if decryption fails.


infeo:
We decided that this is too much refactoring/too intrusive. We can still add the decrypted path later.

PR #291, opened 2025-03-13
Feature: Add timestamp to file system events

infeo:
Relates to #277 

This PR extends the `FilesystemEvent` interface by a `getTimestamp` method and adds a timestamp field to every subclass of `FilesystemEvent`


infeo:
Coverage fails due to the getterMethods (which do not need testing being simple getter methods).

